PR #9, opened 2017-09-07
Changed some variables names to work with the new config file.

CameronRP:
Some variables that I forgot to rename when changing the config file.


CameronRP:
Yeah just missed them.

PR #10, opened 2017-09-07
Added 'where', 'offset', 'limit', and 'Authorization' to the allowed …

CameronRP:
…headers in CORS requests.

This allows the Web Interface to do API calls with these headers.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers


CameronRP:
https://tree.taiga.io/project/the-cacophony-project/us/192?kanban-status=268960

PR #15, opened 2017-09-21
Added a general 'recording' model and file processing API.

CameronRP:
Added a general 'recording' model that will be used instead of the separate Audio/IR/Thermal recording models.
Also started the API for FileProcessing.


api/V1/Recording.js
```diff
@@ -0,0 +1,201 @@
...
+   * - processingMeta
+   *
+   * @apiUse V1DeviceAuthorizationHeader
+   *
+   * @apiParam {JSON} data Metadata about the recording (see above).
+   * @apiParam {File} file File of the recording.
+   *
+   * @apiUse V1ResponseSuccess
+   * @apiuse V1ResponseError
+   */
```
mjs: Nice work remembering to do the documentation up front.

api/V1/Recording.js
```diff
@@ -0,0 +1,201 @@
...
+var AWS = require('aws-sdk');
+
+module.exports = (app, baseUrl) => {
+  var apiUrl = baseUrl + '/recordings';
+
+  /**
+   * @api {post} /api/v1/recordings Add a new recording.
+   * @apiName PostRecording
+   * @apiGroup Recordings
+   * @apiDescription This call is the new way to upload any type of recording.
```
mjs: I don't think this should be described as "the new way". That will be confusing after a few weeks/months.

Just say something like: `This is call is used for uploads of new recordings. It currently supports raw thermal video but will eventually support all recording types.`

api/V1/Recording.js
```diff
@@ -0,0 +1,201 @@
...
+   * It takes a 'data' field which contains a JSON object string that can
+   * contain any of the following fields, Note that 'type' is required:
+   * - type: REQUIRED, one of ('thermalRaw160x120')
```
mjs: I wonder if including the resolution in type too specific. `thermalRaw` might be enough. The upload itself could include the frame dimensions (or it could be in `additionalMetadata`).

api/V1/Recording.js
```diff
@@ -0,0 +1,201 @@
...
+   * @apiuse V1ResponseError
+   */
+  app.post(
+    apiUrl,
+    passport.authenticate(['jwt'], { session: false }),
+    async (request, response) => {
+      log.info(request.method + " Request: " + request.url);
+
+      // Check that the request was authenticated by a device.
+      if (request.user !== null && !requestUtil.isFromADevice(request))
```
mjs: Couldn't `requestUtil.isFromADevice` do the `request.user !== null` check to avoid needing to repeat this everywhere?

api/V1/Recording.js
```diff
@@ -0,0 +1,201 @@
...
+        return responseUtil.notFromADevice(response);
+      var device = request.user;
+
+      var recording;
+      var key = randomstring.generate();
```
mjs: A UUID is standard, commonly used way to give a random, unique ID to something. If it's not too much hassle to change the code to use them, it might be nice to switch to them.

Relevant links:
* https://en.wikipedia.org/wiki/Universally_unique_identifier
* https://www.npmjs.com/package/uuid

api/V1/Recording.js
```diff
@@ -0,0 +1,201 @@
...
+        } catch (e) {
+          log.debug(e);
+          // TODO Stop stream here.
+        }
+      });
+
+      // Stream file to LeoFS.
+      form.on('part', (part) => {
+        if (part.name != 'file') return part.resume();
+        validFile = true;
```
mjs: Should `validFile` only be set to `true` once `s3.upload` has completed successfully?

api/V1/Recording.js
CameronRP: Yeah, that would be better. I guess we don't want too many different "types".

api/V1/Recording.js
CameronRP: Yes, it currently does do that. The reason that I had `request.user !== null` there is so anonymous requests could be done. Saying that I changed it so you can't do anonymous requests so it isn't needed.

api/V1/Recording.js
```diff
@@ -0,0 +1,201 @@
...
+    async (request, response) => {
+      log.info(request.method + " Request: " + request.url);
+
+      // Check that the request was authenticated by a User.
+      if (request.user !== null && !requestUtil.isFromAUser(request))
+        return responseUtil.notFromAUser(response);
+
+      var where = request.query.where;
+      var offset = Math.floor(parseInt(request.query.offset));
+      var limit = Math.floor(parseInt(request.query.limit));
```
mjs: Why is `Math.floor` required here? Isn't the output of `parseInt` already an integer value?

api/fileProcessing/index.js
```diff
@@ -0,0 +1,113 @@
...
+  var apiUrl = '/api/fileProcessing';
+
+  /**
+   * @api {get} /api/fileProcessing Get a new file processing job
+   * @apiName getNewFileProcessingJob
+   * @apiGroup FileProcessing
+   *
+   * @apiParam {String} type Type of recording.
+   * @apiParam {String} state Processing state.
+   */
```
mjs: If the file processing API is internal only, perhaps it shouldn't be included in the docs?

api/fileProcessing/index.js
```diff
@@ -0,0 +1,113 @@
...
+        recording.set('jobKey', null);
+        await recording.save();
+        return response.status(200).json({
+          messages: ["Processing finished."],
+        });
+    } else {
+      recording.set('processingStartTime', null);
+      recording.set('jobKey', null);
+      await recording.save();
+      return response.status(200).json({
```
mjs: Should this be a 5xx error, not 200?

Server.js
```diff
@@ -44,6 +44,13 @@ app.all('*', function(req, res, next) {
 
 require('./api/V1')(app);
 
+// Add file processing API.
+var fileProcessingApp = express();
+fileProcessingApp.use(bodyParser.urlencoded({ extended: false }));
+require('./api/fileProcessing')(fileProcessingApp);
+http.createServer(fileProcessingApp).listen(config.fileProcessing.port);
+log.info('Starting file processing on', config.fileProcessing.port);
```
mjs: What prevents the fileProcessing API from also being served on the normal port?

CameronRP: I had it on a different port so it could only be accessible from the CCL servers.
If we want it open on the normal port we will need to add an authentication process and LeoFS to also be accessible in some way.

api/V1/Recording.js
CameronRP: Yeah, you are right. Not too sure what I was doing..

api/fileProcessing/index.js
CameronRP: The current code for generating the docs doesn't include the fileProcessing directory so it won't be included. But I included the documentation style commenting for self reference.

api/fileProcessing/index.js
CameronRP: This is not a server error but the fileProcessing saying that it failed to process the file. So the status code should be 200?

api/fileProcessing/index.js
mjs: Ok, that's fair enough

PR #17, opened 2017-10-01
Added the ability to download 'Recording' files.

CameronRP:
Also fixed a bug with not accepting users JWTs when getting recordings.

### DIFF
--- a/api/V1/Recording.js
+++ b/api/V1/Recording.js
@@ -9,6 +9,7 @@ var multiparty = require('multiparty');
 var config = require('../../config/config')
 var AWS = require('aws-sdk');
 var uuidv4 = require('uuid/v4');
+var jsonwebtoken = require('jsonwebtoken');
 
 module.exports = (app, baseUrl) => {
   var apiUrl = baseUrl + '/recordings';
@@ -161,7 +162,7 @@ module.exports = (app, baseUrl) => {
       log.info(request.method + " Request: " + request.url);
 
       // Check that the request was authenticated by a User.
-      if (requestUtil.isFromAUser(request))
+      if (!requestUtil.isFromAUser(request))
         return responseUtil.notFromAUser(response);
 
       var where = request.query.where;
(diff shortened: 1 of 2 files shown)


api/V1/Recording.js
```diff
@@ -161,7 +162,7 @@ module.exports = (app, baseUrl) => {
       log.info(request.method + " Request: " + request.url);
 
       // Check that the request was authenticated by a User.
-      if (requestUtil.isFromAUser(request))
+      if (!requestUtil.isFromAUser(request))
```
mjs: oops :)

api/V1/Recording.js
```diff
@@ -216,4 +217,80 @@ module.exports = (app, baseUrl) => {
         rows: result.rows,
       });
     });
+
+    /**
+    * @api {get} /api/v1/recordings/:id Get a recording
+    * @apiName GetRecording
+    * @apiGroup Recordings
+    * @apiDescription This call returns a recording in a JSON format and a JSON
```
mjs: Suggestion: "returns a recording in a JSON format" => "returns metadata for a recording in JSON format"

api/V1/Recording.js
```diff
@@ -216,4 +217,80 @@ module.exports = (app, baseUrl) => {
...
+    * Web Token (JWT) which can be used to retrieve a specific audio recording.
```
mjs: Not just audio right? How about "... can be used to retrieve the recorded content."

api/V1/Recording.js
```diff
@@ -216,4 +217,80 @@ module.exports = (app, baseUrl) => {
...
+        // Query database.
+        var recording = await models.Recording.findOne(query);
+
+        // Check the the processing is finished.
+        if (recording.get('processingState') !== 'FINISHED')
+          return responseUtil.send(response, {
+            status_code: 204,
+            success: false,
+            messages: ["File has not been fully processed yet."],
+          });
```
mjs: Is there a way to retrieve the original upload vs the final processed version via the API? We'll need this for the thermal videos where people may want to look at the processed video, but we'll want the raw CPTV file for training the machine learning model.

CameronRP: I was going to add this at a later date but can add it now if you want? We will need to make sure that you can only get raw data from approved data types also so will need to add the audio recordings types if we add it now.

api/V1/Recording.js
CameronRP: Yes! I'll change that..

mjs:
Looks great.

PR #22, opened 2017-10-12
Change thermal file type to mp4

CameronRP:
Change thermal data to be processed to mp4, added a recording name generated from the date and MIME type, and made the file processing to be able to update the recordings fields (can set MIME type by using this).

### DIFF
--- a/models/Recording.js
+++ b/models/Recording.js
@@ -90,7 +90,7 @@ var apiSettableFields = [
 ];
 
 var processingStates = {
-  thermalRaw: ['toOggVideo', 'FINISHED'],
+  thermalRaw: ['toMp4', 'FINISHED'],
 }
 
 var processingAttributes = [
--- a/models/Recording.js
+++ b/models/Recording.js
@@ -1,7 +1,6 @@
 var util = require('./util/util');
 var validation = require('./util/validation');
 var moment = require('moment-timezone');
-moment().tz("Pacific/Auckland").format();
 
 module.exports = function(sequelize, DataTypes) {
   var name = 'Recording';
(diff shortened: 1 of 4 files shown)


models/Recording.js
```diff
@@ -58,10 +59,19 @@ function canGetRaw() {
   return false;
 }
 
+function getFileName() {
+  var ext = "";
+  if (this.fileMimeType == 'video/mp4') ext = ".mp4"
+  var dateStr = new Date(this.recordingDateTime)
+    .toLocaleString().replace(" ", "T").replace(/:/g, "-");
```
mjs: Is the output from `toLocaleString` dependent on the system locale? If so, this isn't going to be all that reliable.

A more robust approach could be to use moment.js or datejs for parsing and formatting the date.

models/Recording.js
```diff
@@ -58,10 +60,18 @@ function canGetRaw() {
   return false;
 }
 
+function getFileName() {
+  var ext = "";
+  if (this.fileMimeType == 'video/mp4') ext = ".mp4"
+  return moment(new Date(this.recordingDateTime)).tz("Pacific/Auckland")
```
mjs: What date type is `this.recordingDateTime`? Is it already a `Date` meaning there's no need to do the `new Date(...)`?

If it's a string, then I think `moment()` will parse it directly, so again there's no need for the `new Date(...)`.

CameronRP: this.recordingDateTime is a string formatted to the ISO8601 standard.
So if `this.recordingDate` is null then `moment(this.recordingDate).format(...)` will return "Invalid". But with `moment(new Date(this.recordingDate)).format(...)` if `this.recordingDate` is null then it gets the date as 1970...

mjs: Cool - makes sense. As long as it's there for a reason. 
Land away!

PR #25, opened 2017-10-20
Added device to the results of a recording get query.

CameronRP:


api/V1/Recording.js
```diff
@@ -205,6 +205,7 @@ module.exports = (app, baseUrl) => {
         include: [
           { model: models.Group },
           { model: models.Tag },
+          { model: models.Device, where: {}, attributes: ["devicename", "id"] },
```
mjs: Is the where necessary if it's empty?

CameronRP: Without the `where: {}` i get an `SequelizeDatabaseError: column Recording.DeviceId does not exist` when doing an query. Not exactly sure why this is..

PR #27, opened 2017-10-25
Added ability to tag 'Recording' models.

CameronRP:
This also removed the ability to tag older model types, but this wasn't used or going to be used.

### DIFF
--- a/models/Recording.js
+++ b/models/Recording.js
@@ -72,7 +72,6 @@ function getUserPermissions(user) {
   var recording = this;
   return new Promise(async (resolve, reject) => {
     var groupIds = await user.getGroupsIds();
-    console.log(recording.GroupId);
     if (groupIds.indexOf(recording.GroupId) !== -1) {
       permissions.canDelete = true;
       permissions.canTag = true;
--- a/models/Recording.js
+++ b/models/Recording.js
@@ -1,6 +1,7 @@
 var util = require('./util/util');
 var validation = require('./util/validation');
 var moment = require('moment-timezone');
+var models = require('./');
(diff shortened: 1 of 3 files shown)


models/Recording.js
```diff
@@ -48,12 +49,38 @@ module.exports = function(sequelize, DataTypes) {
...
       getFileName: getFileName,
+      getUserPermissions: getUserPermissions,
     },
   };
 
   return sequelize.define(name, attributes, options);
 }
 
+/**
+ * Returns a JSON describing what the user can do to the recording.
```
mjs: `a JSON` should probably just be `JSON`

models/Recording.js
```diff
@@ -48,12 +49,38 @@ module.exports = function(sequelize, DataTypes) {
...
+  // permission. This will be changed in the future.
+  var permissions = {
+    canDelete: false,
+    canTag: false,
+    canView: false,
+  }
+  var recording = this;
+  return new Promise(async (resolve, reject) => {
+    var groupIds = await user.getGroupsIds();
+    if (groupIds.indexOf(recording.GroupId) !== -1) {
```
mjs: What does it mean when the group id isn't in `groupsIds` and is it wise to give full permissions then?

CameronRP: It will only give full permissions if the recording belongs to a group that the user is in. This can be changed in the future to limited permissions when we add custom group permissions.

mjs: Sorry, I read it as "give permissions if the user isn't in the group" but that's obviously not it.

PR #30, opened 2017-11-16
Added devices 'GET' request.

CameronRP:


api/V1/Device.js
```diff
@@ -75,4 +78,40 @@ module.exports = function(app, baseUrl) {
...
+
+          // Check that the request was authenticated by a User.
+          if (!requestUtil.isFromAUser(request))
+            return responseUtil.notFromAUser(response);
+
+          var userGroupIds = await request.user.getGroupsIds();
+          var devices = await models.Device.findAndCount({
+              where: { GroupId: { "$in": userGroupIds } },
+              attributes: ["devicename", "id"],
+          });
```
mjs: ordering?

api/V1/Device.js
```diff
@@ -75,4 +78,40 @@ module.exports = function(app, baseUrl) {
...
+   * @apiUse V1UserAuthorizationHeader
+   *
+   * @apiUse V1ResponseSuccess
+   * @apiSuccess {JSON} devices List of devices.
+   *
+   * @apiUse V1ResponseError
+   */
+  app.get(
+      apiUrl,
+      passport.authenticate(['jwt'], {session: false}),
```
mjs: just so I know, what does the `session: false` here do?

api/V1/Device.js
```diff
@@ -75,4 +78,40 @@ module.exports = function(app, baseUrl) {
...
+      async (request, response) => {
+          log.info(request.method + " Request: " + request.url);
```
mjs: I see that lines like this are everywhere. It seems like there should be some higher level handler logging these somewhere rather than each handler do this.

Not for this PR anyway...

api/V1/Device.js
CameronRP: By device id or name?

api/V1/Device.js
CameronRP: Yep, also need to look at a higher level handler for errors.

api/V1/Device.js
CameronRP: It doesn't look like it actually does anything.
https://github.com/themikenicholson/passport-jwt/blob/master/lib/strategy.js#L90
But it was used in the example so probably best to leave it there for now
https://github.com/themikenicholson/passport-jwt#authenticate-requests

api/V1/Device.js
mjs: Name I guess? The user shouldn't need to know or care about device ids.

PR #34, opened 2017-11-22
Added ability to get recordings are are or aren't tagged.

CameronRP:


api/V1/Recording.js
```diff
@@ -199,19 +199,31 @@ module.exports = (app, baseUrl) => {
         });
       }
 
+      // If query should be filtered by tagged or not or ignored.
+      var sqlLiteral = '';
+      var tagRequired = false;
+      if (where._tagged == true) {
```
mjs: how does `where._tagged` get its value from?

api/V1/Recording.js
```diff
@@ -199,19 +199,31 @@ module.exports = (app, baseUrl) => {
...
+        tagRequired = true;
+      } else if (where._tagged == false) {
+        sqlLiteral = 'NOT EXISTS (SELECT * FROM   "Tags" WHERE  "Tags".id = "Recording".id)';
+        tagRequired = false;
+      }
```
mjs: Looks like instead of messing around with `tagRequired` you can add a `having` section to the query. Something like this might work:

For the doesn't have a tag case:
```
having: ['COUNT(Tags.id) == ?', 0]
```
For the must have a tag case:
```
having: ['COUNT(Tags.id) >= ?', 1]
```

api/V1/Recording.js
CameronRP: From the users query. The web UI will add _tagged = true or false or not add it depending on what radio button is selected.

PR #35, opened 2017-11-22
Allow users to have direct access to devices

mjs:
In some cases it's useful to be able to give a user access to a device's recordings when using group membership is cumbersome. This change introduces a new DeviceUsers table which allows users to access devices. The pre-existing (unused)  user(1) : device(M) association has been removed.

This change also moves all the recording query logic into the model where it should be. This makes it easier to share query functionality between the query and "just get one" aspects.

Still to do:
- [x] Fix the tagged/not tagged bug 
- [x] Update the code that retrieves the devices for the dropdown to include devices accessible via DeviceUsers

### DIFF
--- a/models/Device.js
+++ b/models/Device.js
@@ -30,10 +30,26 @@ module.exports = function(sequelize, DataTypes) {
     },
   };
 
+  var allForUser = async function(user) {
+    var deviceIds = await user.getDeviceIds();
+    var userGroupIds = await user.getGroupsIds();
+    console.log("deviceIds", deviceIds);
+    console.log("userGroupIds", userGroupIds);
+    return this.findAndCount({
+      where: { "$or": [
+        {GroupId: {"$in": userGroupIds}},
+        {id: {"$in": deviceIds}},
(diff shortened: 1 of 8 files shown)


models/Device.js
```diff
@@ -30,10 +30,26 @@ module.exports = function(sequelize, DataTypes) {
     },
   };
 
+  var allForUser = async function(user) {
+    var deviceIds = await user.getDeviceIds();
+    var userGroupIds = await user.getGroupsIds();
+    console.log("deviceIds", deviceIds);
+    console.log("userGroupIds", userGroupIds);
```
CameronRP: We dont't want to use console.log do we?

models/Device.js
```diff
@@ -32,10 +30,26 @@ module.exports = function(sequelize, DataTypes) {
...
```
CameronRP: We don't want to use `console.log()` do we?

mjs: No we don't. That's from when I was checking something.

PR #36, opened 2017-11-23
Adding delete and patch

CameronRP:
Added DELETE, and PATCH for "recordings".
Added new "comment" column to "recordings". Will be able to write a comment for recordings on the view page.

### DIFF
--- a/api/V1/Recording.js
+++ b/api/V1/Recording.js
@@ -282,4 +282,44 @@ module.exports = (app, baseUrl) => {
         });
       }
     );
+
+  /**
+  * @api {delete} /api/v1/recordings/:id Delete an existing recording
+  * @apiName DeleteRecording
+  * @apiGroup Recordings
+  *
+  * @apiUse V1UserAuthorizationHeader
+  *
+  * @apiUse V1ResponseSuccess
+  * @apiUse V1ResponseError
+  */
+  app.delete(
+    apiUrl + '/:id',
+    passport.authenticate(['jwt'], { session: false }),
+    async (request, response) => {
+      log.info(request.method + " Request: " + request.url);
+
+      if (!requestUtil.isFromAUser(request))
+        return responseUtil.notFromAUser(response);
+
+      var id = parseInt(request.params.id);
+      if (!id)
+        return responseUtil.invalidDataId(response);
+
+      var deleted = await models.Recording.deleteOne(request.user, id);
+      if (deleted) {
+        responseUtil.send(response, {
+          statusCode: 200,
+          success: true,
+          messages: ["Deleted recording."],
+        });
+      } else {
+        responseUtil.send(response, {
+          statusCode: 400,
+          success: false,
+          messages: ["Failed to delete recording."],
+        });
+      }
+    }
+  );
 };
--- a/api/V1/Recording.js
+++ b/api/V1/Recording.js
@@ -332,6 +332,7 @@ module.exports = (app, baseUrl) => {
   * If any other values are requested to be updated the request will fail and no
   * updates will occur.
   * - location
+  * - comment
   *
   * @apiUse V1UserAuthorizationHeader
   *
--- a/api/V1/Recording.js
+++ b/api/V1/Recording.js
@@ -322,4 +322,61 @@ module.exports = (app, baseUrl) => {
       }
     }
   );
+
+  /**
+  * @api {patch} /api/v1/recordings/:id Update an existing recording
+  * @apiName UpdateRecording
+  * @apiGroup Recordings
+  * @apiDescription This call is used for updating fields of current recordings.
+  * The only values that can be updated at the moment are the following ones.
+  * If any other values are requested to be updated the request will fail and no
+  * updates will occur.
+  * - location
+  *
+  * @apiUse V1UserAuthorizationHeader
+  *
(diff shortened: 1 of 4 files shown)


api/V1/Recording.js
```diff
@@ -282,4 +282,102 @@ module.exports = (app, baseUrl) => {
...
+        });
+      }
+    }
+  );
+
+  /**
+  * @api {patch} /api/v1/recordings/:id Update an existing recording
+  * @apiName UpdateRecording
+  * @apiGroup Recordings
+  * @apiDescription This call is used for updating fields of current recordings.
```
mjs: replace `of current recordings` with `a previously submitted recording`

api/V1/Recording.js
```diff
@@ -282,4 +282,102 @@ module.exports = (app, baseUrl) => {
...
+  * The only values that can be updated at the moment are the following ones.
+  * If any other values are requested to be updated the request will fail and no
+  * updates will occur.
+  * - location
+  * - comment
```
mjs: Something like this might be better:

The following fields may be updated are:
- location
- comment

If a change to any other field is attempted the request will fail and no update will occur.

api/V1/Recording.js
```diff
@@ -282,4 +282,102 @@ module.exports = (app, baseUrl) => {
...
+  *
+  * @apiUse V1UserAuthorizationHeader
+  *
+  * @apiParam {JSON} updates Fields to update. If any keys are not reconised or can't be updated the whole request will fail and no updates will occur.
```
mjs: There's probably no need to repeat the bit about the request failing if it's just been mentioned.

mjs: Are the fields passed as a list? That could be clearer.

api/V1/Recording.js
```diff
@@ -282,4 +282,102 @@ module.exports = (app, baseUrl) => {
...
+      var updated = await models.Recording.updateOne(request.user, id, updates);
+      if (updated) {
+        return responseUtil.send(response, {
+          statusCode: 200,
+          success: true,
+          messages: ['Updated recording.']
+        });
+      } else {
+        return responseUtil.send(response, {
+          statusCode: 200,
```
mjs: this should be 400 right?

models/Recording.js
```diff
@@ -121,10 +157,13 @@ module.exports = function(sequelize, DataTypes) {
       addAssociations: addAssociations,
       query: query,
       getOne: getOne,
+      deleteOne: deleteOne,
+      updateOne: updateOne,
       processingAttributes: processingAttributes,
       processingStates: processingStates,
       apiSettableFields: apiSettableFields,
       userGetAttributes: userGetAttributes,
+      apiUpdatableFields: apiUpdatableFields,
```
mjs: there's probably no need to export this is there? It's only used here.

Really, I don't think any field names should be exported from the models.

models/Recording.js
```diff
@@ -106,6 +107,41 @@ module.exports = function(sequelize, DataTypes) {
...
+   * Deletes a single recording if the user has permission to do so.
+   */
+  var deleteOne = async function(user, id) {
+    var recording = await this.getOne(user, id);
+    if (recording == null)
+      return false;
+    var userPermissions = await recording.getUserPermissions(user);
+    if (userPermissions.canDelete != true)
+      return false;
+    else {
```
mjs: Not using curly brackets is technically fine but it can end up leading to hard to find bugs.

https://stackoverflow.com/a/4797368/195383

A lot of places have a style guide which says you must always used braces even when not necessary and I tend to lean that way too.

models/Recording.js
CameronRP: Yeah, probably not needed.

models/Recording.js
CameronRP: I am guessing we can/should add that to the .eslintrc.yml file then?

api/V1/Recording.js
CameronRP: It parses the JSON using the key as the column and value as the new value. How would you have a list organized? A list of tuples with [[key1, val1], [key2, val2]]?

api/V1/Recording.js
CameronRP: Yep

api/V1/Recording.js
mjs: Ah of course, it's the fields and new values so a dict/object is more appropriate. `JSON` doens't automatically mean "object" by the way. It could be a JSON array.

Anyway, maybe change the documentation to say: "Object containing the fields to update and their new values." or something.

api/V1/Recording.js
mjs: Fixed.

api/V1/Recording.js
mjs: Fixed.

api/V1/Recording.js
mjs: Fixed.

models/Recording.js
mjs: Yes, but let's do that separately

models/Recording.js
mjs: Fixed

models/Recording.js
mjs: Actually, this was dead easy to do so I added it :)

mjs:
This was finished off in #37

PR #40, opened 2017-12-14
User & group management APIs

### LINKED ISSUE #39: Raw CPTV file downloads need a reasonable filename
When MP4 files are downloaded you get a nicely named file based on the timestamp but CPTV files are just called "rawData". It would be great if CPTV files could use the same naming scheme.

CameronRP:
closes #39 
API added/changed to help with group admin on web:
- GET /api/v1/groups 
- POST /api/v1/groups/users
- DELETE /api/v1/group/users
- GET /api/v1/users This was changed to take a sequelize 'where' statement so you can get users ID, as users IDs are used when adding, removing users from groups.

Added 'superuser' to users. Adds ability to add/remove users from any group.

### DIFF
--- a/api/V1/Group.js
+++ b/api/V1/Group.js
@@ -78,7 +78,7 @@ module.exports = function(app, baseUrl) {
    * @apiGroup Group
    *
    * @apiParam {JSON} where [Sequelize where conditions](http://docs.sequelizejs.com/manual/tutorial/querying.html#where) for query.
-   * @apiParam {Number} userId Only get groups that this user belongs to.
+   * @apiParam {Number} [userId] Only get groups that this user belongs to.
    *
    * @apiUse V1ResponseSuccess
    *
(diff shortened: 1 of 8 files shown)


api/V1/Group.js
```diff
@@ -69,4 +71,154 @@ module.exports = function(app, baseUrl) {
...
+      var groups = await models.Group.findAll({
+        where: where,
+        include: [
+          {
+            model: models.User,
+            attributes: ['id', 'username'],
+            where: userWhere,
+          },
+        ],
+      });
```
mjs: Up to you whether you change it now or later, but I really do think that all DB queries should be kept in the `models` package so that API handler methods are very thin. They should just parse parameters, call to a model method and then structure the return value.

models/GroupUsers.js
```diff
@@ -8,9 +8,28 @@ module.exports = function(sequelize, DataTypes) {
...
+        GroupId: groupId,
+        UserId: userId,
+        admin: true,
+      }
+    });
+    if (groupUsers == null) {
+      return false;
+    } else {
+      return true;
+    }
```
mjs: This if block can just be replaced with `return groupUsers != null;`.

models/User.js
```diff
@@ -22,19 +22,37 @@ module.exports = function(sequelize, DataTypes) {
       type: DataTypes.STRING,
       allowNull: false,
     },
+    superuser: {
+      type: DataTypes.BOOLEAN,
+      defaultValue: false,
+    }
+  };
+
+  const publicFields = Object.freeze([
```
mjs: Good use of Object.freeze :)

PR #41, opened 2017-12-21
Middleware example.

CameronRP:
Just as example of how we can setup the middleware. Not intended to be merged.

### DIFF
--- a/api/V1/Device.js
+++ b/api/V1/Device.js
@@ -6,6 +6,9 @@ var responseUtil = require('./responseUtil');
 var passport = require('passport');
 var log = require('../../logging');
 var requestUtil = require('./requestUtil');
+const middleware = require('../middleware');
+const { check, oneOf, validationResult } = require('express-validator/check');
+const { matchedData, sanitize } = require('express-validator/filter');
 
 module.exports = function(app, baseUrl) {
   var apiUrl = baseUrl + '/devices';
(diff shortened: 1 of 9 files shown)


api/V1/Recording.js
```diff
@@ -155,60 +156,41 @@ module.exports = (app, baseUrl) => {
...
+    middleware.logging,
+    middleware.parseParams({
+      query: {
+        where: { type: 'JSON' },
+        offset: { type: 'INTEGER' },
+        limit: { type: 'INTEGER' },
+        order: { type: 'JSON', required: false },
+      },
+    }),
+    middleware.authenticate('user'),
```
mjs: This looks great!

api/V1/User.js
```diff
@@ -24,38 +25,55 @@ module.exports = function(app, baseUrl) {
...
-        success: false,
-        messages: ['Missing username or password.']
+  app.post(
+    apiUrl,
+    middleware.logging,
+    middleware.parseParams({
+      body: {
+        username: { type: 'STRING' },
+        password: { type: 'STRING' },
+      },
```
mjs: Rather than writing our code to do parameter validation what about using this? https://www.npmjs.com/package/express-validator

There's a number of other similar packages, but this looks like one of the best ones to me (based only a quick look though)

api/middleware.js
```diff
@@ -0,0 +1,166 @@
...
+  return device;
+};
+
+/**
+ * Wraps the function in a Promise and will properly pass errors to next.
+ */
+const asyncWrapper = fn => (request, response, next) => {
+  Promise.resolve(fn(request, response, next))
+    .catch(next);
+};
```
mjs: Nice

api/middleware.js
```diff
@@ -0,0 +1,166 @@
...
+        });
+      }
+    } catch(err) {
+      next(err);
+    }
+    next();
+  };
+};
+
+const parseJWTUser = async (jwtDecoded) =>  {
```
mjs: Nit: this isn't really parsing anything... how about `lookupUser` or something?

api/middleware.js
```diff
@@ -0,0 +1,166 @@
...
+  const user = await models.User.findById(jwtDecoded.id);
+  if (user == null) {
+    const err = new Error();
+    err.message = 'user not found from JWT';
+    throw err;
+  }
+  return user;
+};
+
+const parseJWTDevice = async (jwtDecoded) => {
```
mjs: As above

api/V1/Device.js
```diff
@@ -92,22 +77,22 @@ module.exports = function(app, baseUrl) {
...
-      async (request, response) => {
-          log.info(request.method + " Request: " + request.url);
-
-          // Check that the request was authenticated by a User.
-          if (!requestUtil.isFromAUser(request))
-            return responseUtil.notFromAUser(response);
+    apiUrl, [
+      check('Authorization').custom((...args) => {
+        return util.authenticate('user', ...args);
+      })
```
mjs: Presumably we can wrap this up into a function so that we have one helper for APIs that require users and one for APIs that require devices.

mjs:
This PR can probably be closed now right?

clare:
Yes...sorry if I didn't say that.

mjs:
@clare This is @CameronRP's PR not yours :)

mjs:
This has been implemented.

PR #42, opened 2018-01-04
Add support for automatic tags

mjs:
Tags now have an `automatic` flag which indicates that the tag was generated by an automatic classifier.

Recordings can now be selected according to any of the following criteria:
- 'any'
- 'untagged'
- 'tagged'
- 'automatic-only'
- 'human-only'
- 'automatic+human'

Also fixed up many eslint issues.

### DIFF
--- a/models/Tag.js
+++ b/models/Tag.js
@@ -33,6 +33,9 @@ module.exports = function(sequelize, DataTypes) {
     age: { // Guessed age in weeks of animal
       type: DataTypes.INTEGER
     },
+    automatic: { //
+      type: DataTypes.BOOLEAN
+    },
   };
 
   var options = {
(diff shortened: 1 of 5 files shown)


models/Tag.js
```diff
@@ -33,6 +33,9 @@ module.exports = function(sequelize, DataTypes) {
     age: { // Guessed age in weeks of animal
       type: DataTypes.INTEGER
     },
+    automatic: { //
+      type: DataTypes.BOOLEAN
```
CameronRP: I think we should also include the default value and allowNull here so it's the same as it was defined in the migration.

mjs: I wasn't sure if those properties were meant to be here as well so I left them out.

mjs: Of course it makes sense to have it though (for when starting from scratch?). Fixed.

PR #43, opened 2018-01-05
Added DeviceUsers API.

CameronRP:
Added ability to add and delete users from devices.
Changed GET API for devices.

### DIFF
--- a/api/V1/Device.js
+++ b/api/V1/Device.js
@@ -82,32 +82,138 @@ module.exports = function(app, baseUrl) {
   /**
    * @api {get} /api/v1/devices Get list of devices
    * @apiName GetDevices
-   * @apiGroup Devices
+   * @apiGroup Device
    *
-   * @apiUse V1UserAuthorizationHeader
+   * @apiParam {JSON} where [Sequelize where conditions](http://docs.sequelizejs.com/manual/tutorial/querying.html#where) for query.
+   * @apiParam {Number} [userId] Get devices that belong to this user.
    *
    * @apiUse V1ResponseSuccess
-   * @apiSuccess {JSON} devices List of devices.
    *
    * @apiUse V1ResponseError
    */
   app.get(
-      apiUrl,
-      passport.authenticate(['jwt'], {session: false}),
-      async (request, response) => {
-          log.info(request.method + " Request: " + request.url);
+    apiUrl,
+    async (request, response) => {
+      log.info(request.method + " Request: " + request.url);
 
-          // Check that the request was authenticated by a User.
-          if (!requestUtil.isFromAUser(request))
-            return responseUtil.notFromAUser(response);
+      var where = request.query.where;
+      var queryUserId = request.query.userId;
+      try {
+        where = JSON.parse(where);
+      } catch (e) {
+        return responseUtil.send(response, {
+          statusCode: 400,
+          success: false,
+          messages: ['Could not parse "where" to a JSON'],
+        });
+      }
 
-          var devices = await models.Device.allForUser(request.user);
+      var devices = await models.Device.query(where, queryUserId);
 
-          return responseUtil.send(response, {
-              devices: devices,
-              statusCode: 200,
-              success: true,
-              messages: ["completed get devices query"],
-          })
+      return responseUtil.send(response, {
+        devices: devices,
+        statusCode: 200,
+        success: true,
+        messages: ["completed get devices query"],
       });
+    }
+  );
+
+  /**
+  * @api {post} /api/v1/devices/users Add a user to a device.
+  * @apiName AddUserToDevice
+  * @apiGroup Device
+  * @apiDescription This call can add a user to a device. This is so individuals
+  * can monitor there devices without being part of the group that the device
+  * belongs to.
+  *
+  * @apiUse V1UserAuthorizationHeader
+  *
+  * @apiParam {Number} deviceId ID of the device.
+  * @apiParam {Number} userId ID of the user to add to the device.
+  * @apiParam {Boolean} admin If the user should be an admin for the device.
+  *
+  * @apiUse V1ResponseSuccess
+  * @apiUse V1ResponseError
+  */
+  app.post(
+    apiUrl + '/users',
+    passport.authenticate(['jwt'], { session: false }),
+    async function(request, response) {
+      log.info(request.method + ' Request: ' + request.url);
+
+      if (!requestUtil.isFromAUser(request)) {
+        return responseUtil.notFromAUser(response);
+      }
+
+      var added = await models.Device.addUserToDevice(
+        request.user,
+        request.body.deviceId,
+        request.body.userId,
+        request.body.admin,
+      );
+
+      if (added) {
+        return responseUtil.send(response, {
+          statusCode: 200,
+          success: true,
+          messages: ['Added user to device'],
+        });
+      } else {
+        return responseUtil.send(response, {
+          statusCode: 400,
+          success: false,
+          messages: ['failed to add user to device']
+        });
+      }
+    }
+  );
+
+  /**
+  * @api {delete} /api/v1/devices/users Removes a user from a device.
+  * @apiName RemoveUserFromDevice
+  * @apiGroup Device
+  * @apiDescription This call can remove a user from a deive. Has to be
+  * authenticated by an admin from the group that the device belongs to or a
+  * user that has contorl of device.
+  *
+  * @apiUse V1UserAuthorizationHeader
+  *
+  * @apiParam {Number} userId ID of the user to delete from the device.
+  * @apiParam {Number} deviceId ID of the device.
+  *
+  * @apiUse V1ResponseSuccess
+  * @apiUse V1ResponseError
+  */
+  app.delete(
+    apiUrl + '/users',
+    passport.authenticate(['jwt'], { session: false }),
+    async function(request, response) {
+      log.info(request.method + ' Request: ' + request.url);
+
+      if (!requestUtil.isFromAUser(request)) {
+        return responseUtil.notFromAUser(response);
+      }
+
+      var removed = await models.Device.removeUserFromDevice(
+        request.user,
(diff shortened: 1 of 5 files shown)


api/V1/Device.js
```diff
@@ -82,32 +82,138 @@ module.exports = function(app, baseUrl) {
...
+    async (request, response) => {
+      log.info(request.method + " Request: " + request.url);
 
-          // Check that the request was authenticated by a User.
-          if (!requestUtil.isFromAUser(request))
-            return responseUtil.notFromAUser(response);
+      var where = request.query.where;
+      var queryUserId = request.query.userId;
+      try {
+        where = JSON.parse(where);
```
mjs: I'm not crazy about accepting arbitrary where clauses from the API as it could be abused.

What do you actually need to be able to search by and could there just be a specific API for that use?

api/V1/Device.js
```diff
@@ -82,32 +82,138 @@ module.exports = function(app, baseUrl) {
...
+        return responseUtil.send(response, {
+          statusCode: 200,
+          success: true,
+          messages: ['Added user to device'],
+        });
+      } else {
+        return responseUtil.send(response, {
+          statusCode: 400,
+          success: false,
+          messages: ['failed to add user to device']
```
mjs: The code is inconsistent on capitalisation used in these messages. We should pick a policy and stick to it.

models/Device.js
```diff
@@ -30,25 +30,110 @@ module.exports = function(sequelize, DataTypes) {
...
-    var userGroupIds = await user.getGroupsIds();
-    return this.findAndCount({
-      where: { "$or": [
-        {GroupId: {"$in": userGroupIds}},
-        {id: {"$in": deviceIds}},
-      ]},
+  const models = sequelize.models;
+
+  /**
+  * Adds/update a user to a Device, if the given user has permissino to do so.
```
mjs: typo: permissino

api/V1/Device.js
```diff
@@ -82,32 +82,138 @@ module.exports = function(app, baseUrl) {
...
+      if (!requestUtil.isFromAUser(request)) {
+        return responseUtil.notFromAUser(response);
+      }
+
+      var added = await models.Device.addUserToDevice(
+        request.user,
+        request.body.deviceId,
+        request.body.userId,
+        request.body.admin,
+      );
```
mjs: We really need to get that error handling middleware in place so that if this fails due to for example, PostgreSQL not being up, a 500 error gets returned.

api/V1/Device.js
```diff
@@ -82,32 +82,138 @@ module.exports = function(app, baseUrl) {
...
+        });
+      }
+    }
+  );
+
+  /**
+  * @api {delete} /api/v1/devices/users Removes a user from a device.
+  * @apiName RemoveUserFromDevice
+  * @apiGroup Device
+  * @apiDescription This call can remove a user from a deive. Has to be
```
mjs: typo: deive

api/V1/Device.js
```diff
@@ -82,32 +82,138 @@ module.exports = function(app, baseUrl) {
...
+  * authenticated by an admin from the group that the device belongs to or a
+  * user that has contorl of device.
```
mjs: typo: contorl

api/V1/Device.js
```diff
@@ -82,32 +82,138 @@ module.exports = function(app, baseUrl) {
...
+      if (removed) {
+        return responseUtil.send(response, {
+          statusCode: 200,
+          success: true,
+          messages: ['Removed user from the device'],
+        });
+      } else {
+        return responseUtil.send(response, {
+          statusCode: 400,
+          success: false,
```
mjs: I've always wondered, why do we have a `success` boolean as well as the HTTP status code to indicate success/failure? Shouldn't we just use the HTTP status?

Not for this PR obviously... just wondering.

models/Device.js
```diff
@@ -30,25 +30,110 @@ module.exports = function(sequelize, DataTypes) {
...
+    const userToAdd = await models.User.findById(userToAddId);
+    if (device == null || userToAdd == null) {
+      return false;
+    }
+    // Return false if the user doesn't have permission.
+    const isGroupAdmin = await models.GroupUsers.isAdmin(device.groupId, authUser.id);
+    const isDeviceAdmin = await models.DeviceUsers.isAdmin(device.id, authUser.id);
+    if (!authUser.superuser && !isGroupAdmin && !isDeviceAdmin) {
+      return false;
+    }
```
mjs: The logic at the top of this function is clear and well organised. Nice.

models/Device.js
```diff
@@ -30,25 +30,110 @@ module.exports = function(sequelize, DataTypes) {
...
+
+    // Get association if already there and update it.
+    var deviceUser = await models.DeviceUsers.findOne({
+      where: {
+        DeviceId: deviceId,
+        UserId: userToAdd.id,
+      }
+    });
```
mjs: Instead of doing this direct query, I wonder if you could access the deviceUsers association via a property on the `device` that's already been loaded at the top of the function. sequelize adds properties to model objects for each association.

models/Device.js
```diff
@@ -30,25 +30,110 @@ module.exports = function(sequelize, DataTypes) {
...
+    const userToRemove = await models.User.findById(userToRemoveId);
+    if (device == null || userToRemove == null) {
+      return false;
+    }
+    // Return false if the user doesn't have permission.
+    const isGroupAdmin = await models.GroupUsers.isAdmin(device.groupId, authUser.id);
+    const isDeviceAdmin = await models.DeviceUsers.isAdmin(device.id, authUser.id);
+    if (!authUser.superuser && !isGroupAdmin && !isDeviceAdmin) {
+      return false;
+    }
```
mjs: Can these permission checks be extracted to a separate function and shared with `addUserToDevice`?

models/Device.js
```diff
@@ -30,25 +30,110 @@ module.exports = function(sequelize, DataTypes) {
...
+
+    // Check that association is there to delete.
+    const deviceUsers = await models.DeviceUsers.findAll({
+      where: {
+        DeviceId: device.id,
+        UserId: userToRemove.id,
+      }
+    });
```
mjs: As above

models/Device.js
```diff
@@ -30,25 +30,110 @@ module.exports = function(sequelize, DataTypes) {
...
+    if (deviceUsers.length == 0) {
+      return false;
```
mjs: This should probably return `true`. Typically it's not considered an error when a "remove" API finds that something has already been removed.

You could just remove this if block and let the for loop run (it won't when there's nothing to remove).

api/V1/Device.js
CameronRP: How do you think this could be abused? Should we have it that you can only see devices that you have control of?

api/V1/Device.js
CameronRP: Yeah, I didn't bother adding in proper error handling here for the reason that we will be adding that middleware soon, hopefully.

api/V1/Device.js
CameronRP: Should we use no capitalisation like Go?

api/V1/Device.js
CameronRP: Yeah, it's not really needed, or used. Might as well remove it.

models/Device.js
CameronRP: I think so. If you say to include Users in the device query.

models/Device.js
CameronRP: OK, I think that will need to be changed in the GroupUsers API also.

api/V1/Device.js
mjs: Sure, I'm fine with that.

api/V1/Device.js
mjs: I've created an issue to sort this out: https://github.com/TheCacophonyProject/cacophony-api/issues/48

api/V1/Device.js
mjs: Allowing arbitrary external input to be used as part of the query greatly increases the chances that someone will be able to get at data they're not supposed to be to. We're also then trusting that sequelize has no bugs which allow SQL injection attacks.

Allowing sequelize input also ties us to using sequelize. If we wanted to change how the API server accesses the database or be written in a different language we would have to break compatibility and figure out how to handle existing clients which use sequelize query. 

Details of the database layer shouldn't be leaking out to the API layer. It's safer & cleaner if we have specific query arguments for the things we need to expose over the API.

api/V1/Device.js
mjs: Cool, I'm fine with that.

models/Device.js
mjs: Ok. Up to you whether you leave it as is or go via the association.

models/Device.js
mjs: OK. Please do that as well.

models/Group.js
```diff
@@ -102,6 +93,26 @@ module.exports = function(sequelize, DataTypes) {
...
+      console.log(permissions);
+      return permissions;
+    }
+
+    if (await models.GroupUsers.isAdmin(this.id, user.id)) {
+      permissions.canAddUsers = true;
+      permissions.canRemoveUsers = true;
+    }
+    return permissions;
+  };
```
mjs: a similar cleanup as above can be made here

models/User.js
```diff
@@ -40,6 +42,17 @@ module.exports = function(sequelize, DataTypes) {
...
 
+  const getGroupDeviceIds = async function() {
+    var groupIds = await this.getGroupsIds();
+    var devices = await models.Device.findAll({
+      where: { GroupId: { "$in": groupIds }},
+      attributes: ['id'],
+    });
+    var idList = [];
+    for (var key in devices) { idList.push(devices[key].id); }
+    return idList;
```
mjs: A shorter and nicer way of doing this is: 
```js
return devices.map(d => d.id);
```

PR #54, opened 2018-01-27
using express validator with Device API

CameronRP:
Started to implement express-validator.

### DIFF
--- a/api/V1/Device.js
+++ b/api/V1/Device.js
@@ -1,10 +1,9 @@
-var models = require('../../models');
-var jwt = require('jsonwebtoken');
-var config = require('../../config/config');
-var responseUtil = require('./responseUtil');
-var passport = require('passport');
-var log = require('../../logging');
-var requestUtil = require('./requestUtil');
+const models       = require('../../models');
+const jwt          = require('jsonwebtoken');
+const config       = require('../../config/config');
+const responseUtil = require('./responseUtil');
+const middleware   = require('../middleware');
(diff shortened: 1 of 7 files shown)


models/User.js
```diff
@@ -51,11 +51,21 @@ module.exports = function(sequelize, DataTypes) {
     return devices.map(d => d.id);
   };
 
+  const getFromId = async function(id) {
+    return await this.findById(id);
+  };
+
+  const getFromName = async function(name) {
```
mjs: Just curious, how is `const foo = async function(...` different from `async function foo(...`?

api/V1/Device.js
```diff
@@ -23,60 +22,29 @@ module.exports = function(app, baseUrl) {
...
+        .custom(value => { return models.Device.freeDevicename(value); }),
+      middleware.checkNewPassword('password'),
+      middleware.getGroup,
+    ],
+    middleware.requestWrapper(async (request, response) => {
+      const device = await models.Device.create({
+        devicename: request.body.devicename,
+        password: request.body.password,
+        GroupId: request.body.group.id,
       });
```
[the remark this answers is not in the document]
mjs: With the index in place, the middelware check isn't really necessary.

models/User.js
CameronRP: With `async function foo(...` you can reassign foo to something else, `foo = 1` and with `const` you can't.

PR #58, opened 2018-02-08
Adding more express validator

CameronRP:
Added express validator to more of the API. Also some small changes to code style.

### DIFF
--- a/api/V1/AuthenticateDevice.js
+++ b/api/V1/AuthenticateDevice.js
@@ -1,11 +1,11 @@
-var models = require('../../models');
-var jwt = require('jsonwebtoken');
-var config = require('../../config/config');
-var util = require('./util');
-var responseUtil = require('./responseUtil');
+const jwt          = require('jsonwebtoken');
+const config       = require('../../config/config');
+const responseUtil = require('./responseUtil');
+const middleware   = require('../middleware');
+const { body }     = require('express-validator/check');
 
 module.exports = function(app) {
-  /** 
+  /**
   * @api {post} /authenticate_device/ Authenticate a device
   * @apiName AuthenticateDevice
   * @apiGroup Authentication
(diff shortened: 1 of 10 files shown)


api/V1/AuthenticateDevice.js
```diff
@@ -17,55 +17,31 @@ module.exports = function(app) {
...
-      });
-  });
+      const passwordMatch = await request.body.device.comparePassword(request.body.password);
+      if (passwordMatch) {
+        var data = request.body.device.getJwtDataValues();
+        var token = 'JWT ' + jwt.sign(data, config.server.passportSecret);
+        return responseUtil.send(response, {
+          statusCode: 200,
+          success: true,
+          messages: ["Successfull login."],
```
mjs: Successful

mjs: Also, I think we were going to avoid the full stops at the end of messages?

api/V1/AuthenticateDevice.js
```diff
@@ -17,55 +17,31 @@ module.exports = function(app) {
...
+          token: token
+        });
+      } else {
+        return responseUtil.send(response, {
+          statusCode: 401,
+          success: false,
+          messages: ["Wrong password or devicename."]
```
mjs: ditto

api/V1/Group.js
```diff
@@ -26,56 +21,24 @@ module.exports = function(app, baseUrl) {
...
-            success: false,
-            messages: [err.message]
-          });
-        } else {
-          responseUtil.serverError(res, err);
-        }
+  app.post(
+    apiUrl,
+    [
+      middleware.authenticateUser,
```
mjs: So much nicer

api/V1/Group.js
```diff
@@ -142,21 +90,20 @@ module.exports = function(app, baseUrl) {
...
-
-      if (!requestUtil.isFromAUser(request)) {
-        return responseUtil.notFromAUser(response);
-      }
+    [
+      middleware.authenticateUser,
+      middleware.getGroup,
+      middleware.getUser,
+      check('admin').isBoolean(),
+    ],
```
mjs: Nice!

api/V1/AuthenticateDevice.js
CameronRP: Yeah, I didn't go through changing all the messages on this commit. But will do that later.

PR #60, opened 2018-02-21
Document the tagMode and tags query args of the Recordings API

mjs:
The docs were missed when the args were added.


mjs:
Good suggestion. Done.

PR #64, opened 2018-02-26
/api/V1/users -> /api/V1/users/:username

mjs:
Replace the API which returns all the users matching some criteria with one which returns the details for just one username. The latter is the all the web UI actually wants.

It's also good to avoid APIs which take raw sequelize strings as are security risks as and limit our implementation flexibility.

The web UI is being updated to switch to this new API.

### DIFF
--- a/api/middleware.js
+++ b/api/middleware.js
@@ -68,7 +68,7 @@ const getModelById = function(modelType, fieldName, checkFunc=check) {
     if (model === null) {
       throw new Error(format('could not find %s of %s', fieldName, val));
     }
-    req.body[attrName(modelType)] = model;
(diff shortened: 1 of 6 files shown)


api/middleware.js
```diff
@@ -62,41 +62,49 @@ const signedUrl = query('jwt').custom((value, {req}) => {
...
+function attrName(modelType) {
+  switch (modelType) {
+  case models.User:
+    return "user";
+  case models.Group:
+    return "group";
+  case models.Device:
+    return "device";
+  default:
+    throw "unknown model type: " + modelType;
```
CameronRP: Try getting name from Sequelize model here.

PR #65, opened 2018-03-15
Api test

clare:
First cut of the the api-tests. 

The tests are not testing much right now but its a start.


Server.js
```diff
@@ -6,18 +6,21 @@ var winston = require('winston');
...
-}
-var config = require('./config/config');
-var models = require('./models');
-var log = require('./logging');
+var configPath = './config'
+process.argv.forEach((val, index) => {
+  if (val.toLowerCase().startsWith('config=')) {
+    configPath  = val.split('=')[1];
+  }
+});
```
mjs: The convention for Unix-ish command line args is for multi-character options to start with `--` (i.e. `--config=`). This should really also fail if unsupported options are passed.

Instead of doing this all manually, try one of the many NodeJS packages available for command line processing. For simple command line option handling (like here) the following look promising:
* https://www.npmjs.com/package/minimist
* https://github.com/yargs/yargs
* https://github.com/sindresorhus/meow

Server.js
```diff
@@ -6,18 +6,21 @@ var winston = require('winston');
...
+
+var allConfig = require('./cacconfig');
```
mjs: Can `cacconfig` just be called `config`?

Also, why assign the imported cacconfig module to a variable named `allConfig`? That's a bit confusing.

Server.js
```diff
@@ -6,18 +6,21 @@ var winston = require('winston');
...
+allConfig.loadConfig(configPath);
+var config = allConfig.config;
```
mjs: `loadConfig` should should attach the top level configuration fields it loads from the configuration file as attributes of the `cacconfig` module. This would avoid the need to do `allConfig.config` everywhere. 

This will just be a straightforward for loop of the config object created from parsing the configuration file.

cacconfig.js
```diff
@@ -0,0 +1,27 @@
...
+    console.log(`Looking for config files in directory "${path.resolve(configDirPath)}"`);
+
+    checkConfigFileExists(configDirPath, 'config.js');
+    exports.config = require(configDirPath + '/config');
+      
+    checkConfigFileExists(configDirPath, 'database.js');
+    exports.database = require(configDirPath + '/database');
+}
+
+function checkConfigFileExists(configDir, filename) {
```
mjs: Thanks for providing some nicer handling for missing configuration files.

models/index.js
```diff
@@ -3,7 +3,8 @@ var path = require('path');
 var Sequelize = require('sequelize');
 var basename = path.basename(module.filename);
 var env = process.env.NODE_ENV || 'development';
-var config = require(__dirname + '/../config/database.js')[env];
+var config = require('../cacconfig').database[env];
```
mjs: Perhaps call this `dbConfig` to make it clear it's not the same as `config` everywhere else?

test/README.md
```diff
@@ -0,0 +1,52 @@
...
+# To run all tests
+
+* Create a virtualenv using your preferred method.
+* Install dependencies: `pip install -r requirements.txt`
+* Run with: `pytest -s`
+
+
+# To create a separate server for testing
+
+Create a new minio server
```
mjs: Include download link for Minio?

test/helper.py
```diff
@@ -0,0 +1,93 @@
...
+            self._admin = UserAPI(self.config.api_server, self.config.admin_username, self.config.admin_password).login()
+        return self._admin
+
+    def _print_actual_name(self, name):
+        print("  ({})".format(name))
+
+    def _check_admin_exists(self):
+        try:
+            self._get_admin()
+        except Exception:
```
mjs: Please look for a more specific exception to catch here. As it stands, if **anything** goes wrong in `_get_admin()` this code will try to create the admin user, even if the exception has nothing at all to do with the admin user not being there.

Instead of "looking before you leap" another approach is just to try and create the admin and allow for it to fail because the admin already exists. I'm not sure that can be easily distinguished from other errors from the API but if it can be, that seems like a cleaner solution.

test/test_03_group.py
```diff
@@ -0,0 +1,46 @@
...
+        daniel.cannot_see_any_recordings()
+
+        # print("  and Daniel shouldn't be able to add himself to clare's group")
+        # daniel.add_to_group(daniel, claresGroup)
+
+        # print("When Clare adds Daniel to her group")
+        # clare.add_to_group(daniel, claresGroup)
+
+        # print("  then Daniel should see the recording from 'Terminator'")
+        # clare.can_see_recording_from(terminator)
```
mjs: Not finished?

test/userapi.py
```diff
@@ -0,0 +1,91 @@
...
+        url = urljoin(self._baseurl, "/api/v1/users/{}".format(username))
+        response = requests.get(url, headers=self._auth_header)
+        response.raise_for_status()
+        # print(response.json())
+        
+
+# def add_user_to_group(self, username, groupname) 
+    #     url = urljoin(self._baseurl, "/api/v1/groups")
+    #     response = requests.post(url, data={'groupname': groupname}, headers=self._auth_header)
+    #     response.raise_for_status()
```
mjs: remove this for now?

clare:
Pull request merged

PR #67, opened 2018-03-21
Search test

clare:
Added a test that tests get by tag, and get by tag type.

### DIFF
--- a/test/testrecording.py
+++ b/test/testrecording.py
@@ -1,0 +1,26 @@
+class TestRecording:
+
+    def __init__(self, recordingId):
+        self.recordingId = recordingId
+
+    def is_tagged_as(self, animal):
+        return TestTagPromise(self, animal)
+        
+
+class TestTagPromise():
+    def __init__(self, testrecording, animal):
+        self._testrecording = testrecording
+        self._tagData = {"animal": animal}
+        print("  which is tagged as a {}".format(animal), end='')
+
+    def by(self, testuser):
+        print(" by {}".format(testuser.username))
+        testuser.tag_recording(self._testrecording.recordingId, self._tagData)
+        return self._testrecording
+
+    def byAI(self, testuser):
+        print(" by {}".format("by AI"))
+        self._tagData['automatic'] = True
+
+        testuser.tag_recording(self._testrecording.recordingId, self._tagData)
+        return self._testrecording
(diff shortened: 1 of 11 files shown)


test/testrecording.py
```diff
@@ -0,0 +1,26 @@
...
+        self._testrecording = testrecording
+        self._tagData = {"animal": animal}
+        print("  which is tagged as a {}".format(animal), end='')
+
+    def by(self, testuser):
+        print(" by {}".format(testuser.username))
+        testuser.tag_recording(self._testrecording.recordingId, self._tagData)
+        return self._testrecording
+
+    def byAI(self, testuser):
```
mjs: It does this take `testuser`? It doesn't use it.

Maybe call this method `is_automatic()` or something?

test/testuser.py
```diff
@@ -6,18 +6,59 @@ def __init__(self, username, userapi):
...
-        recordings = self._userapi.query(limit=1)
+    def when_searching_with_tagmode(self, tagmode):
+        queryParams = {"tagmode": tagmode}
+        return RecordingQueryPromise(self, queryParams)
+
+    def when_searching_for_tags(self, *tags):
+        queryParams = {"tags": tags}
+        return RecordingQueryPromise(self, queryParams)
+
+    def can_see_recordings(self, *expectedTestRecordings):
```
mjs: The convention in python is to use lower_case_with_underscores for variables names, not camelCase.

test/userapi.py
```diff
@@ -84,6 +84,11 @@ def get_user_details(self, username):
         response.raise_for_status()
         # print(response.json())
         
+    def tag_recording(self, recordingId, tagDictionary):
+        url = urljoin(self._baseurl, "/api/v1/tags/")
+        tagData = {"tag": json.dumps(tagDictionary), "recordingId": recordingId}
+        response = requests.post(url, headers=self._auth_header, data=tagData)
+        response.raise_for_status()
```
mjs: use check_response instead?

test/testrecording.py
clare: The code for doing this is all in userapi, so needs a user.   Without a user you have to somehow get the URL etc so I just took a short cut and get the caller to pass in the admin user.   I now it's not strictly necessary.   Feel free to change it.

PR #68, opened 2018-03-22
Log to console only

### LINKED ISSUE #53: Log to stdout
Instead of managing its own log file the API server should just log to stdout and let systemd manage the log output. This is better aligned with modern practices things and avoids issues with managing log file permissions and rotation.

mjs:
Let systemd manage the log output instead of having cacophony-api manage its own log files. This also avoids the confusion of having different log levels spread across different files.

Sequelize's logs are now also sent through winston (at debug). This allows them to be turned off by selecting a log level of info or higher.

A default log level is now set so that database migrations work again.

Closes #53.

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,6 +1,12 @@
 var path = require('path');
 var fs = require('fs');
 
+// Set some default configuration
+var server = {
+  loggerLevel: "info",
+};
+
+
 function loadConfig(configDirPath) {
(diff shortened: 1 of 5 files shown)


clare:
Code looks ok.   Merge away!

mjs:
@clare you close the PR which meant it didn't merge!

PR #69, opened 2018-03-23
Add initial tests for AudioRecordings API

mjs:
These test that audio can be uploaded and is visible on the server. More to come.


mjs:
Combining this with more changes in another PR

PR #73, opened 2018-03-23
The --config argument now takes a file rather than a directory

mjs:
Now that there's only one configuration file it doesn't need to take a directory.

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,5 +1,5 @@
-var path = require('path');
-var fs = require('fs');
+const fs = require('fs');
+const path = require('path');
 
 // Set some default configuration
 var server = {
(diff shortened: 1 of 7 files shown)


test/README.md
```diff
@@ -36,13 +36,13 @@ Create a new database called 'cacophonytest'
   - `\q`
 
 Migrate the database to the latest version
```
clare: Actually maybe we should name the test config file testapp.js.   That would make sure that we are getting the right file.

config.js
```diff
@@ -1,45 +1,38 @@
...
-    for (var key in config) {
-        exports[key] = config[key];
-    }
+  checkDatabaseConfigAvailable(config);
 
-    testDatabaseConfigAvailable(config);
-
-    return exports;
+  return exports;
 }
```
clare: Might be worth using path.resolve(configDirPath) to get the full path here for this error.   That way user can see exactly where the program is looking.

mjs: Already done. The first thing that loadConfig() does is path.resolve(configPath).

test/README.md
mjs: Yep, that's a good idea. Eventually I think we want the generation of the test config file to be automated but using a more obvious configuration file name now is a good idea.

I'll add that to this PR.

config.js
clare: Sorry missed that.

Issue #33, opened 2017-11-16
Fix errors reported by eslint


mjs:
There's a lot of errors reported. Some are minor (but worth sorting out) - some look more serious.

PR #74, opened 2018-03-23
Return recording id when adding an audio recording

mjs:
Also add initial test for AudioRecordings API.


api/V1/util.js
```diff
@@ -67,14 +67,15 @@ function addRecordingFromPost(model, request, response) {
 
       modelInstance.set('DeviceId', device.id);
       modelInstance.set('GroupId', device.GroupId);
-      if (typeof device.public == 'boolean')
+      if (typeof device.public == 'boolean') {
```
clare: I could start a fight about the lack of brackets...I prefer brackets for one line statements because it means if someone adds a second line it doesn't break.   This could especially be a problem since we are working with python at the same time...

test/test_02_device.py
```diff
@@ -1,23 +1,33 @@
...
 
-        print("And the CPTV file should be visible to super users on the site")
-        helper.admin_user().can_see_recordings(recording)
+        print("And 'Destroyer' should be able to upload a CPTV file")
+        destroyer.upload_recording()
 
+        print("And the CPTV file should be visible to super users")
+        helper.admin_user().can_see_recording_from(destroyer)
 
+    def test_device_can_upload_audio(self):
```
clare: Not a biggy but I would be tempted to put it in a different file with a different filename.  Since one is a cell phone device and one is a thermal camera device.  Keep the tests based around user scenarios.

test/test_02_device.py
```diff
@@ -1,23 +1,33 @@
...
+
+        print("Then 'Listener' should able to log in")
+        helper.login_as_device(listener.devicename)
+
+        print("And 'Listener' should be able to upload an audio file")
+        recording = listener.upload_audio_recording()
 
-    
\ No newline at end of file
+        print("And the CPTV file should be visible to super users")
```
clare: And CPTV -> Audio file

api/V1/util.js
mjs: I completely agree with you! We should always use curly brackets in JS even when not strictly required. In fact our eslint config even warns about this (we'll start running eslint automatically on pull requests soon once the code is in slightly better shape).

See:
* https://github.com/TheCacophonyProject/cacophony-api/blob/master/.eslintrc.yml#L16
* https://github.com/TheCacophonyProject/cacophony-api/issues/33

I didn't actually touch this line myself. I suspect my editor just fixed up trailing whitespace.

test/test_02_device.py
mjs: Sounds good. I'll split them up.

test/test_02_device.py
mjs: Hooray for Friday afternoon copy and paste :)

api/V1/util.js
clare: Sorry I got it round the wrong way!

PR #75, opened 2018-03-26
Introduce "helper" fixture

mjs:
This makes it easier to make the Helper class available to tests. Using the "module" scope means that the helper is only instantiated once per test module.

Relevant docs: https://docs.pytest.org/en/latest/fixture.html


mjs:
yeah pytest has a lot of magical features. I'm normally wary of too much magic but pytest's magic is well thought out and helps to make tests much more readable.

PR #77, opened 2018-03-29
Adding tests

CameronRP:


test/test_01_user.py
```diff
@@ -12,3 +12,11 @@ def test_can_create_new_user(self, helper):
         userdetails = bob.get_user_details(helper.admin_user())
 
         print("Bob's user id is {}".format(userdetails))
+
+    def test_can_download_recording(self, helper):
+        device = helper.given_new_device(self, 'device')
+        recording = device.has_recording()
+
+        print("\nA user should be able to downlad the recording")
```
mjs: download

test/userapi.py
```diff
@@ -101,6 +101,12 @@ def download_audio(self, recording_id):
         d = self._check_response(r)
         return self._download_signed(d['jwt'])
 
+    def download_recording(self, recording_id):
```
mjs: This doubles up on functionality already on this class

mjs:
Very similar tests which cover the same functionality were added as part of #81 .

PR #81, opened 2018-04-09
More tests

mjs:
This adds tests to check the downloaded content and metadata for audio and recordings, as well as deletion of recordings.

A few API documentation issues have also been fixed.

### DIFF
--- a/test/test_07_user_thermal.py
+++ b/test/test_07_user_thermal.py
@@ -1,13 +1,13 @@
 class TestUserThermal:
-    def test_can_download_audio(self, helper):
-        device = helper.given_new_device(self, 'cacophonator')
+    def test_can_download_recording(self, helper):
+        device = helper.given_new_device(self, 'cacophonator-download')
         recording = device.has_recording()
 
         print("\nA user should be able to download the recording")
         helper.admin_user().can_download_correct_recording(recording)
 
     def test_can_delete_recording(self, helper):
-        device = helper.given_new_device(self, 'cacophonator')
+        device = helper.given_new_device(self, 'cacophonator-delete')
         recording = device.has_recording()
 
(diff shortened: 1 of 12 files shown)


test/testuser.py
```diff
@@ -81,6 +81,42 @@ def cannot_see_any_recordings(self):
                 "User '{}' can see a recording from '{}'".format(
                     self.username, recordings[0]['Device']['devicename']))
 
+    def can_download_correct_recording(self, recording):
+        content = io.BytesIO()
```
clare: This is a very 'programic way to test this'   So it tests this but it 
a) doesn't clarify very well what is tested
b) and would be easy to get the two props arrays confused so you get a test that can't fail.  
It's fine, but maybe we should talk about it on Thursday.

test/test_07_user_thermal.py
```diff
@@ -0,0 +1,21 @@
+class TestUserThermal:
+    def test_can_download_audio(self, helper):
+        device = helper.given_new_device(self, 'cacophonator')
+        recording = device.has_recording()
+
+        print("\nA user should be able to download the recording")
+        helper.admin_user().can_download_correct_recording(recording)
+
+    def test_can_delete_recording(self, helper):
+        device = helper.given_new_device(self, 'cacophonator')
```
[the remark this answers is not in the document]
mjs: OK, will do.

test/testuser.py
mjs: I quite like having a test helper which checks that a recording can be downloaded and has the correct metadata. Happy to discuss other approaches on Thursday (putting it on my list).

mjs:
@clare aren't tags already covered by the tests you added in test_04_tagging.py?

PR #85, opened 2018-05-04
Schedule api

clare:
Finished the schedule api now.   I haven't added any checking for the format of the schedule though.   

I will now move onto the web interface.

### DIFF
--- a/api/V1/Schedule.js
+++ b/api/V1/Schedule.js
@@ -50,6 +50,7 @@ module.exports = (app, baseUrl) => {
 
       var instance = models.Schedule.build(request.body, ["schedule"]);
       instance.set('UserId', request.user.id);
+      // TODO make the device and scedule changes apply in a single transaction
       await instance.save();
 
       await models.Device.update(
@@ -60,7 +61,7 @@ module.exports = (app, baseUrl) => {
       return responseUtil.send(response, {
         statusCode: 200,
         success: true,
-        messages: ['Added new schedule.'],
+        messages: ['Added new schedule for the calling device(s).'],
       });
     })
   );
@@ -113,6 +114,7 @@ module.exports = (app, baseUrl) => {
         await request.user.checkUserControlsDevices([device.id]);
       }
       catch (error) {
+        // TODO this should probably be in the normal requestWrapper
         if (error.name == 'UnauthorizedDeviceException') {
           return responseUtil.send(response, {
             statusCode: 400,
@@ -130,14 +132,18 @@ module.exports = (app, baseUrl) => {
 };
 
 async function getSchedule(device, response, user = null) {
-  var schedule = (device.ScheduleId) ? await models.Schedule.findById(device.ScheduleId) : {schedule: {}};
-  if (!schedule) {
-    return responseUtil.send(response, {
-      statusCode: 400,
-      success: false,
-      devicename: device.devicename,
-      messages: ["Cannot find schedule."],
-    });
+  var schedule = {schedule: {}};
+
+  if (device.ScheduleId) {
+    schedule = await models.Schedule.findById(device.ScheduleId);
+    if (!schedule) {
+      return responseUtil.send(response, {
+        statusCode: 400,
+        success: false,
+        devicename: device.devicename,
+        messages: ["Cannot find schedule."],
+       });
+    }
   }
 
   // get all the users devices that are also associated with this same schedule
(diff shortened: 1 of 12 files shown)


api/V1/Schedule.js
```diff
@@ -0,0 +1,151 @@
...
+   * @apiName PostSchedule
+   * @apiGroup Schedules
+   * @apiDescription This call is used to upload a new audio bait schedule which controls when
+   * sound bait files are played.
+   *
+   * @apiUse V1UserAuthorizationHeader
+   *
+   * @apiBody {JSON} data Metadata about the schedule in JSON format.
+   *
+   * @apiParam {Int[]} devices List of device Ids that schedule should apply to
```
mjs: ids

api/V1/Schedule.js
```diff
@@ -0,0 +1,151 @@
...
+        if (error.name == 'UnauthorizedDeviceException') {
+          return responseUtil.send(response, {
+            statusCode: 400,
+            success: false,
+            messages: [error.message]
+          });
+        } else {
+          throw error;
+        }
+      }
```
mjs: Should this be another middleware helper? At the very least it should be a separate function in this file so that the GET API can share this functionality.

api/V1/Schedule.js
```diff
@@ -0,0 +1,151 @@
...
+
+      var instance = models.Schedule.build(request.body, ["schedule"]);
+      instance.set('UserId', request.user.id);
+      await instance.save();
+
+      await models.Device.update(
+        {ScheduleId: instance.id},
+        {where: {id: deviceIds}}
+      );
```
mjs: It would be better if the insertion of the schedule and the device schedule updates were done in a single database transaction.

api/V1/Schedule.js
```diff
@@ -0,0 +1,151 @@
...
+      return responseUtil.send(response, {
+        statusCode: 200,
+        success: true,
+        messages: ['Added new schedule.'],
+      });
+    })
+  );
+
+  /**
+   * @api {get} api/v1/schedules/ Get device audio bait schedule (for this device)
```
mjs: "for the calling device" perhaps?

api/V1/Schedule.js
```diff
@@ -0,0 +1,151 @@
...
+  var schedule = (device.ScheduleId) ? await models.Schedule.findById(device.ScheduleId) : {schedule: {}};
+
+  if (!schedule) {
+    return responseUtil.send(response, {
+      statusCode: 400,
+      success: false,
+      devicename: device.devicename,
+      messages: ["Cannot find schedule."],
+    });
+  }
```
mjs: This would probably be clearer as:

```js
var schedule = {schedule: {}};

if (device.ScheduleId) {
  schedule = await models.Schedule.findById(device.ScheduleId);
  if (!schedule) {
    return responseUtil.send(response, {
      statusCode: 400,
      success: false,
      devicename: device.devicename,
      messages: ["Cannot find schedule."],
     });
  }
}
```

Using the ternary operator here hides that the `if (!schedule)` condition is only every going to trigger if findById is called.

api/middleware.js
```diff
@@ -119,13 +119,13 @@ const getEventDetailById = getModelById(models.EventDetail, 'eventDetailId');
...
 const checkNewName = function(field) {
-  return check(field, 'invalid name')
+  return body(field, 'invalid name')
     .isLength({ min: 3 })
     .matches(/^[a-zA-Z0-9]+(?:[_ -]?[a-zA-Z0-9])*$/);
 };
 
 const checkNewPassword = function(field) {
-  return check(field, 'invalid password')
+  return body(field, 'Invalid password.  Password must be at least 8 characters long.')
```
mjs: Better, thanks

models/User.js
```diff
@@ -162,6 +163,25 @@ function getDeviceIds() {
     });
 }
 
+function UnauthorizedDeviceException(username, deviceId) {
+  this.name = "UnauthorizedDeviceException";
+  this.message = ("Unauthorized use of device " + deviceId + " by " + username);
+}
+
+UnauthorizedDeviceException.prototype = new Error();
+UnauthorizedDeviceException.prototype.constructor = UnauthorizedDeviceException;
```
mjs: Consider using https://www.npmjs.com/package/node-custom-errors

test/test_10_schedule.py
```diff
@@ -0,0 +1,65 @@
+import pytest
+import time
+
+class TestSchedule:
+    def test_only_device_owners_can_set_scedule(self, helper):
```
mjs: schedule

PR #89, opened 2018-05-31
Added custom error handeling

CameronRP:


api/customErrors.js
```diff
@@ -0,0 +1,67 @@
+const log    = require('../logging');
+const format = require('util').format;
+
+function errorHandler(err, request, response, next) { // eslint-disable-line
+  if (err instanceof CustomError) {
+    response.status(err.statusCode).json(err.toJson());
+    return;
+  }
+  log.error(err);
```
mjs: Do you think we should always log the error, even if it's a CustomError?

I guess a CustomError is more controlled/expected ... it might still be nice to have them in the logs though.

Issue #71, opened 2018-03-23
Run tests under TravisCI


mjs:

ben-biddington:
Hey can you confirm all tests are passing locally?

ben-biddington:
I am trying to stub out `openS3` so I am sure I have made a mistake somewhere, but just in case

clare:
Hi Ben, 

Welcome to Cacophony - great to have you helping.  

I'll get the latest and double check for you, but I think they are working right now.   

FYI the tests are not unit tests - they are integration level tests and we normally run them against an instance of server running against Minio (openS3) and a postgres database on our machines.   The instructions are here https://github.com/TheCacophonyProject/cacophony-api/blob/master/test/README.md.  

Have you remembered to set up the super user?   This isn't automated yet and would be a great place to start.   

Otherwise can you give me a better idea of the errors you are getting and I might be able to help more.

FYI A few of the tests do test saving to S3 and getting the files back so that could cause a few issues if it wasn't working but I wouldn't expect all the tests to fail then. 

Cheers
Clare

clare:
Yes, just confirmed.   All tests are passing right now.

ben-biddington:
Draft [here](https://github.com/ben-biddington/cacophony-api/blob/docker/models/util/util.js#L322)

mjs:
Hi @ben-biddington. To be clear, we're not using S3 itself (for the integration tests or in production) but a S3 clone. We're using the excellent [Minio](https://minio.io/downloads.html#download-server-linux-x64) which is very easy to operate (single binary, one command to run).

It *should* be fairly easy to run Minio inside a TravisCI test run.

What we really need is to automate the steps required to set up for running tests both to make things easier on developer machines but also to facilitate running the tests automatically with TravisCI.

mjs:
Eventually it'll also be nice to have a JavaScript based unit tests. Then, a fake storage might be useful.

ben-biddington:
Hmm just seems like a lot of hassle for something not that interesting (file storage). It is also quite tightly stuck where an abstraction could be.

Will give it a go

mjs:
OK, if you're keen then give it a go.

mjs:
It looks like you've made great progress. Thank you! Having the tests running under TravisCI will be a big step forward. A Docker build is nice to have to.

[...]

PR #99, opened 2018-06-12
Avoid double response when audio recording not found

mjs:
When getFileData fails reponseUtil.invalidFileRequest is called and
then an error is thrown. The thrown error is then picked up by the
catch handled which then ends up calling
responseUtil.invalidDatapointGet, resulting in 2 HTTP responses being
sent (not valid).

This is a likely fix for one of the issues reported in #71.

This code is all going away soon but it's good to fix this in the meantime.

### DIFF
--- a/api/V1/util.js
+++ b/api/V1/util.js
@@ -88,25 +88,27 @@ function addRecordingFromPost(model, request, response) {
 }
 
 function getRecordingFile(modelClass, request, response) {
-
   // Check that if the request was authenticated that is was
   // authenticated by a user JWT, not a device JWT.
   // TODO get passport to do this...
-  if (request.user !== null && !requestUtil.isFromAUser(request))
-  {return responseUtil.notFromAUser(response);}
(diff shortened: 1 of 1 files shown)


api/V1/util.js
```diff
@@ -88,22 +88,23 @@ function addRecordingFromPost(model, request, response) {
...
 
 function getRecordingFile(modelClass, request, response) {
-
   // Check that if the request was authenticated that is was
   // authenticated by a user JWT, not a device JWT.
   // TODO get passport to do this...
-  if (request.user !== null && !requestUtil.isFromAUser(request))
-  {return responseUtil.notFromAUser(response);}
+  if (request.user !== null && !requestUtil.isFromAUser(request)) {
+    return responseUtil.notFromAUser(response);
```
CameronRP: Can put the return statement on a new line like you have below.

mjs: Will do.

PR #101, opened 2018-06-12
Added device and username to logs in request.

### LINKED ISSUE #93: Identify user/device id in logs
It's currently harder to troubleshoot issues then it should be because request logs don't include the related user name or device id.

CameronRP:
closes #93

### DIFF
--- a/api/middleware.js
+++ b/api/middleware.js
@@ -162,21 +162,19 @@ const parseArray = function(field) {
 };
 
 const requestWrapper = fn => (request, response, next) => {
+  var logMessage = format('%s %s', request.method, request.url);
   if (request.user) {
-    log.info(format('%s %s (user: %s)'),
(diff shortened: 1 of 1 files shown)


api/middleware.js
```diff
@@ -162,7 +162,21 @@ const parseArray = function(field) {
...
+    );
+  } else if (request.device) {
+    log.info(format('%s %s (device: %s)'),
+      request.method,
+      request.url,
+      request.device.get("devicename"),
+    );
+  } else {
+    log.info(format('%s %s', request.method, request.url));
+  }
```
mjs: Given the `<method> <URL>` part is always the same, how about creating that first and then using it log calls in the user, device, not-logged-in branches?

This makes the code more concise and avoids unintentional differences creeping in as the code evolves.

PR #103, opened 2018-06-13
Allow tests to run on travis CI

ben-biddington:
The changes are for #71

I think there is still a problem somewhere with the audio recording as I sometimes get all passing, other times between 2 and 4 tests failing.


.travis.yml
```diff
@@ -1,6 +1,21 @@
-language: node_js
-node_js:
-  - "8.9.4"
+sudo: required
+
+services:
+  - docker
+  
+dist: trusty
```
mjs: Can this be xenial? We use 16.04 everywhere so would prefer to be consistent here too.

README.md
```diff
@@ -1,5 +1,7 @@
 # cacophony-api
 
+[![Status](https://api.travis-ci.org/TheCacophonyProject/cacophony-api.svg)](https://travis-ci.org/TheCacophonyProject/cacophony-api)
```
mjs: Thanks for remember to do this (I always forget)

config/app_test_default.js
```diff
@@ -0,0 +1,40 @@
+// A real test configuration
+// @todo: consider readinf from env vars
```
mjs: reading

ben-biddington:
Yep, on its way

clare:
Hey Ben, 

I've been a bit busy lately doing the audio stuff but I just wanted to say a big thanks for picking this up.   It is awesome to have our tests running on checkin and getting us started on a dockerised container too. 

Thanks 
Clare :)

ben-biddington:
Not at all @clare this is an awesome project, let me know if I can help with anything else

PR #107, opened 2018-06-25
Docker improvements

mjs:
* Configure PostgreSQL in docker image instead of in the docker entry
  point. There no reason to delay this until container startup.
* Simplify stopping of the container
* Expose PostgreSQL server port from container. This is useful for
  development and confirming DB contents from the
  outside. PostgreSQL's port is exposed at 5400 to avoid collisions
  with a PostgreSQL server on the host.
* Avoid the need for the container pid file Given that a static container name is used, the pid file is unnecessary.
* Move minio startup to docker-entrypoint. This makes more sense given
  that setting up Minio doesn't depend on the cacophony-api
  code. This also speeds things up somewhat.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ apidoc/*
 testconfig.json
 .pytest_cache
 __pycache__
-docker-container-pid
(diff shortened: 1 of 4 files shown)


ben-biddington:
Make it so!

mjs:
@ben-biddington thanks for the review. Can you please submit a review with an approval? The PR can't be merged without a positive review.

ben-biddington:
Sure -- how?

mjs:
@ben-biddington thanks!

PR #119, opened 2018-07-19
Addition of GroupUser details to group query (to enabled group page user management.)

adbs1:


adbs1:
It does look longwinded, but that initial userWhere clause limits the groups to the current users. Removing it means every group is returned for a non-superuser.

I did spend some time trying to modify the original query as I thought it should be possible in one query. I messed with the GroupUser associations and added them as includes, but I could only get a) All groups plus all associated groupUsers or b) Only groups for the current user plus only the current user in the groupUser set. What I want is: Only groups for the current user plus all associated the groupUsers.

Any ideas?

mjs:
Hmmm... not sure why Travis isn't running. Investigating...

PR #129, opened 2018-08-16
Copy AudioRecordings query params from headers to query params

mjs:
recordingUtil.query() only looks in query params which meant the limit, offset and where fields were being ignored.

Also fixed how `where` params are sent to the API server by the tests. The tests where using an undocumented way of doing this which didn't match what the UI uses.

### DIFF
--- a/test/test_06_user_audio.py
+++ b/test/test_06_user_audio.py
@@ -45,6 +45,26 @@ class TestUserAudio:
         print("Clare should be able to see exactly those the recordings")
         clare.can_see_audio_recordings(recordings)
 
+    def test_can_limit_audio_query(self, helper):
+        print("If a new user grace", end='')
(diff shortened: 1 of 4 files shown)


test/test_06_user_audio.py
```diff
@@ -45,6 +45,26 @@ def test_can_query_audio_recordings(self, helper):
         print("Clare should be able to see exactly those the recordings")
         clare.can_see_audio_recordings(recordings)
 
+    def test_can_limit_audio_query(self, helper):
+        print("If a new user grace", end='')
+        clare = helper.given_new_user(self, 'grace')
```
clare: Is your user Clare or grace?

mjs: oops. copy pasta error. will fix.

PR #132, opened 2018-08-31
Added user email

CameronRP:
When registering a user requires an email now.
A user can login using either an email or username.

### DIFF
--- a/migrations/20180830022142-add-index-to-email-migration.js
+++ b/migrations/20180830022142-add-index-to-email-migration.js
@@ -1,0 +1,11 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface) {
+    return queryInterface.addIndex('Users', {fields:['email'], unique:true});
+  },
+
+  down: function (queryInterface) {
+    return queryInterface.removeIndex('Users', ['email']);
+  }
+};
--- a/models/User.js
+++ b/models/User.js
@@ -64,8 +64,8 @@ module.exports = function(sequelize, DataTypes) {
   };
 
   const freeEmail = async function(email) {
+    email = email.toLowerCase();
     var user = await this.findOne({where: {email: email}});
-    console.log(user);
     if (user != null) {
       throw new Error('email in use');
     }
@@ -119,7 +119,8 @@ module.exports = function(sequelize, DataTypes) {
       checkUserControlsDevices: checkUserControlsDevices,
     },
     hooks: {
-      afterValidate: afterValidate
+      beforeValidate: beforeValidate,
+      afterValidate: afterValidate,
     }
   };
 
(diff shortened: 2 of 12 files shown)


migrations/20180830022142-add-index-to-email-migration.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface) {
+    return queryInterface.addIndex('Users', {fields:['email'], unique:true});
```
mjs: As well as ensuring uniqueness this also ensures that lookups by email are fast

models/User.js
```diff
@@ -103,18 +119,20 @@ module.exports = function(sequelize, DataTypes) {
...
+      afterValidate: afterValidate,
     }
   };
 
   // Define table
   return sequelize.define(name, attributes, options);
 };
 
-var apiSettableFields = [
+const apiSettableFields = Object.freeze([
```
mjs: :+1:

models/User.js
```diff
@@ -204,6 +222,14 @@ function afterValidate(user) {
   });
 }
 
+function beforeValidate(user) {
+  return new Promise((resolve) => {
+    console.log(user.getDataValue('email'));
```
mjs: probably don't need the console.log now

test/helper.py
```diff
@@ -19,7 +19,7 @@ def __init__(self):
 
     def login_as(self, username):
         password = self._make_password(username)
-        api = UserAPI(self.config.api_server, username, password).login()
+        api = UserAPI(self.config.api_server, username, None, password).login()
```
mjs: If you wanted to avoid the extra param you could use the existence of a '@' to know whether an email address or username was being used.

test/userapi.py
```diff
@@ -9,8 +9,12 @@
...
-    def __init__(self, baseurl, username, password='password'):
+    def __init__(self, baseurl, username, email, password='password'):
         super().__init__('user', baseurl, username, password)
+        self.email = email
+
+    def register_as_new(self):
+        return super().register_as_new(email = self.email)
 
     def query(self, startDate=None, endDate=None, min_secs=0, limit=100, offset=0, tagmode=None, tags=None):
         where = defaultdict(dict)
```
mjs: Please add a test where the user logs in using their email address.

models/User.js
```diff
@@ -204,6 +222,14 @@ function afterValidate(user) {
   });
 }
 
+function beforeValidate(user) {
```
mjs: good idea to use a sequelize hook for this

models/User.js
```diff
@@ -58,6 +59,19 @@ module.exports = function(sequelize, DataTypes) {
...
   };
 
+  const getFromEmail = async function(email) {
+    return await this.findOne({where: {email: email}});
+  };
+
+  const freeEmail = async function(email) {
+    email = email.toLowerCase();
+    var user = await this.findOne({where: {email: email}});
+    if (user != null) {
```
mjs: `if (user) { ` would be enough right?

test/helper.py
CameronRP: The UserAPI will need the email most of the time so it can register. This time it doesn't need it as it is just logging in.

PR #133, opened 2018-09-01
Api message cleanup

### LINKED ISSUE #59: Change messages/logs to be formatted consistently
Let's go with all lower case as is commonly done for Go projects.

SimonM2000:
Fixes #59 
Also adds license headers for all API files

### DIFF
--- a/api/customErrors.js
+++ b/api/customErrors.js
@@ -1,4 +1,22 @@
-const log    = require('../logging');
+/*
+cacophony-api: The Cacophony Project API server
+Copyright (C) 2018  The Cacophony Project
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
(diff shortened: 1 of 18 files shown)


api/customErrors.js
```diff
@@ -20,7 +38,7 @@ function CustomError() {
 CustomError.prototype = new Error();
 
 function ValidationError(errors) {
-  this.name = "validation";
+  this.name = "Validation";
```
mjs: `this.name` gets used to set `errorType` in the JSON returned for the error. `errorType` is for the client code to examine and so should remain lower case. Can you arrange it so that `errorType` stays lower case please?

api/middleware.js
```diff
@@ -146,7 +164,7 @@ const parseJSON = function(field) {
       req[location][path] = JSON.parse(value);
       return true;
     } catch(e) {
-      throw new Error(format('could not parse field %s to a json', path));
+      throw new Error(format('Could not parse field %s to a json.', path));
```
mjs: `to a json` is an odd construction. How about `to JSON`?
(I know this isn't from you)

api/middleware.js
```diff
@@ -157,7 +175,7 @@ const parseArray = function(field) {
     try {
       arr = JSON.parse(value);
     } catch(e) {
-      throw new Error(format('could not parse field %s to a json', path));
+      throw new Error(format('Could not parse field %s to a json.', path));
```
mjs: as above

api/customErrors.js
SimonM2000: Ok, should I change this.name = "Client" to "client" later in the file (line 77)? Or just leave everything how it was.

On the error toString, i can: hard-code the string ("Validation error[%d]..."), change it to upper-case (this.name.charAt(0).toUpperCase() + this.name.slice(1)) or just let it stay lower-case.

SimonM2000: Done

PR #135, opened 2018-09-06
Tidy up middleware checks.

CameronRP:
This removes the uses of the general `check` function and uses more specific express-validator functions like `param` `header` `body` `query`
https://express-validator.github.io/docs/check-api.html

### DIFF
--- a/api/V1/Group.js
+++ b/api/V1/Group.js
@@ -19,7 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 const models       = require('../../models');
 const responseUtil = require('./responseUtil');
 const middleware   = require('../middleware');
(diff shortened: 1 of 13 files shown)


api/V1/Group.js
```diff
@@ -19,7 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 const models       = require('../../models');
 const responseUtil = require('./responseUtil');
 const middleware   = require('../middleware');
-const { check }    = require('express-validator/check');
+const { check, query, body } = require('express-validator/check');
```
mjs: where is `check` still being used?

CameronRP: Missed one! Will fix that.

PR #140, opened 2018-09-25
Swap superuser field for global permissions

CameronRP:
Have replaced the Users `superuser` field with `globalPermission`
`globalPermission` currently has three states. 
* `off` - no special permissions. 
* `read` - Can see all recordings
* `write` - Can read/write to all recording/groups. Same permissions the superuser use to have.
Also added an admin API to allow admins (users with global `write` permission) to change the globalPermission of other users.


models/User.js
```diff
@@ -23,10 +23,11 @@ module.exports = function(sequelize, DataTypes) {
       type: DataTypes.STRING,
       allowNull: false,
     },
-    superuser: {
-      type: DataTypes.BOOLEAN,
-      defaultValue: false,
-    }
+    globalPermission: {
+      type: DataTypes.ENUM,
+      values: ['off', 'read', 'write'],
```
mjs: Here's that list again :)

There should probably be constants in this file for the 3 possible values (which are just these string values, but this helps guard against typos) and then a const frozen array with the 3 possible values which can be used where the array is currently being hardcoded.

This helps to avoid errors and helps to find the locations where permissions are being referred to.

models/User.js
```diff
@@ -128,6 +131,19 @@ module.exports = function(sequelize, DataTypes) {
   return sequelize.define(name, attributes, options);
 };
 
+const changeGlobalPermission = async function(admin, user, permission) {
+  if (!user || !admin || !admin.isAdmin()) {
+    return false;
+  }
+  user.setDataValue('globalPermission', permission);
```
mjs: I'm pretty sure (but not 100%) that you can just do: `user.globalPermission = permission;`

models/User.js
```diff
@@ -190,7 +206,7 @@ UnauthorizedDeviceException.prototype = new Error();
 UnauthorizedDeviceException.prototype.constructor = UnauthorizedDeviceException;
 
 const checkUserControlsDevices = async function(deviceIds) {
-  if (!this.superuser) {
+  if (this.globalPermission != 'write') {
```
mjs: `if (!this.isAdmin())`?

test/test_11_global_permissions.py
```diff
@@ -0,0 +1,95 @@
+import pytest
+
+class TestGlobalPermission:
+
+    def test_setting_global_permissions(self, helper):
+        print("Given new users Tom and bob")
+        tom = helper.given_new_user(self, 'tom')
+        bob = helper.given_new_user(self, 'bob')
+
+        print("  Tom shoudl not be able to change global permissions for himself")
```
mjs: should

test/test_11_global_permissions.py
```diff
@@ -0,0 +1,95 @@
...
+
+        print("  Or other users (bob)")
+        with pytest.raises(OSError):
+            tom.set_global_permission(bob.username, 'write')
+
+        print("When Tom is given global read permission")
+        admin = helper.admin_user()
+        admin.set_global_permission(tom.username, 'read')
+
+        print(" he shoudl still not be able to change global permissions for himself")
```
mjs: should

test/test_11_global_permissions.py
```diff
@@ -0,0 +1,95 @@
...
+        fred.can_see_recordings(recording)
+
+        print("  And update the recording")
+        fred.update_recording(recording, comment='testing')
+
+        print("  And tag the recording")
+        fred.tag_recording(recording.recordingId, {"animal":'cat'})
+
+        print("  And delete the recording")
+        fred.delete_recording(recording)
```
mjs: great tests

models/User.js
```diff
@@ -128,6 +131,19 @@ module.exports = function(sequelize, DataTypes) {
...
+    return false;
+  }
+  user.setDataValue('globalPermission', permission);
+  await user.save();
+  return true;
+};
+
+const isAdmin = function() {
+  return this.globalPermission == 'write';
+};
```
mjs: You could replace this with a more general `hasPermission` and replace all the permission checks with calls to this.

models/Device.js
```diff
@@ -168,7 +168,7 @@ module.exports = function(sequelize, DataTypes) {
   };
 
   Device.userPermissions = async function(user) {
-    if (user.superuser) {
+    if (user.isAdmin()) {
```
mjs: It might be better to call this `hasGlobalWrite` to be consistent with `hasGlobalRead`. This makes the meaning clearer too ("admin" is a bit vague).

models/User.js
```diff
@@ -20,6 +20,15 @@ var bcrypt = require('bcrypt');
...
 const Op = Sequelize.Op;
 
+const PERMISSION_WRITE = 'write';
+const PERMISSION_READ = 'read';
+const PERMISSION_OFF = 'off';
+const PERMISSIONS = Object.freeze([
+  PERMISSION_WRITE,
+  PERMISSION_READ,
+  PERMISSION_OFF,
+]);
```
mjs: Great

models/User.js
```diff
@@ -63,13 +73,15 @@ module.exports = function(sequelize, DataTypes) {
...
   ]);
-  
+
   User.apiSettableFields = Object.freeze([
     'firstName',
     'lastName',
     'email'
   ]);
 
+  User.globalPermissions = PERMISSIONS;
```
mjs: Is it possible to make `globalPermissions` a const to ensure it doesn't get accidentally reassigned (not sure if you can, just wondering)

mjs: Should it be `GLOBAL_PERMISSIONS` because it's a constant value?

PR #142, opened 2018-09-28
Added filter options for recordings query.

CameronRP:
Recording location coordinates are filter to an accuracy of 100m by default.

### DIFF
--- a/api/V1/Recording.js
+++ b/api/V1/Recording.js
@@ -119,10 +119,7 @@ module.exports = (app, baseUrl) => {
    * <li>human-only: match only recordings which have been manually tagged
    * <li>automatic+human: match only recordings which have been both automatically & manually tagged
    * </ul>
-   * @apiParam {JSON} [filterOptions] options for filtering the recordings data.
-   * <ul>
-   * <li>latLongAcc: Maximum accuracy of latitude longitude coordinates in meters. Minimum 100m
-   * </ul>
+   * @apiUse FilterOptions
    *
    * @apiUse V1ResponseSuccessQuery
    *
@@ -164,6 +161,7 @@ module.exports = (app, baseUrl) => {
    * @apiUse MetaDataAndJWT
    * @apiUse V1UserAuthorizationHeader
    *
+   * @apiUse FilterOptions
    * @apiUse V1ResponseSuccess
    * @apiSuccess {String} downloadFileJWT JSON Web Token to use to download the
    * recording file.
@@ -178,6 +176,7 @@ module.exports = (app, baseUrl) => {
     [
       middleware.authenticateUser,
       param('id').isInt(),
+      middleware.parseJSON('filterOptions', query).optional(),
     ],
     middleware.requestWrapper(async (request, response) => {
       const { recording, rawJWT, cookedJWT } = await recordingUtil.get(request);
--- a/api/V1/Recording.js
+++ b/api/V1/Recording.js
@@ -119,6 +119,10 @@ module.exports = (app, baseUrl) => {
    * <li>human-only: match only recordings which have been manually tagged
    * <li>automatic+human: match only recordings which have been both automatically & manually tagged
    * </ul>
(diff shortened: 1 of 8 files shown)


models/Recording.js
```diff
@@ -115,9 +115,43 @@ module.exports = function(sequelize, DataTypes) {
...
+    } else {
+      latLong[0] -= resolution/2;
+    }
+    latLong[1] = latLong[1] - latLong[1]%resolution;
+    if (latLong[1] > 0) {
+      latLong[1] += resolution/2;
+    } else {
+      latLong[1] -= resolution/2;
+    }
+  }
```
mjs: How about:

```js

function reduceLatLonPrecision(latLon, prec) {
  assert (latLon.length == 2);

  const resolution = prec * 360 / 40000000;
  const half_resolution = resolution / 2;
  return latLon.map((val) => {
     val = val - (val % resolution);
     if (val > 0) {
        val += half_resolution;
     } else {
        val -= half_resolution;
     }
     return val;
  });
}
```
This avoid duplicating the code for lat and lon (less chance of bugs) and returning a new array instead of modifying the existing one in place reduces the chances for strange bugs (side effects are evil).

The call above would then change to:
```js
recordings[i].location.coordinates = reduceLatLonPrecision(recordings[i].location.coordinates, options.latLongAcc);
```

models/Recording.js
```diff
@@ -115,9 +115,43 @@ module.exports = function(sequelize, DataTypes) {
...
     };
-    return this.findAndCount(q);
+
+    var queryResponse = await this.findAndCount(q);
+    filterRecordings(user, queryResponse.rows, filterOptions);
+    return queryResponse;
   };
 
+  function filterRecordings(user, recordings, options = {}) {
+    if (typeof options.latLongAcc != 'number') {
```
mjs: "Prec" vs "Acc" here too

models/Recording.js
```diff
@@ -115,9 +115,43 @@ module.exports = function(sequelize, DataTypes) {
...
+      options.latLongAcc = 100;
+    }
+    if (!user.hasGlobalWrite()) {
+      options.latLongAcc = Math.max(options.latLongAcc, 100);
+    }
+
+    for (var i in recordings) {
+      if (recordings[i].location) {
```
mjs: suggest you start the loop with `const rec = recordings[i];` to avoid having to index into recordings all the time.

api/V1/Recording.js
```diff
@@ -119,6 +119,10 @@ module.exports = (app, baseUrl) => {
    * <li>human-only: match only recordings which have been manually tagged
    * <li>automatic+human: match only recordings which have been both automatically & manually tagged
    * </ul>
+   * @apiParam {JSON} [filterOptions] options for filtering the recordings data.
+   * <ul>
+   * <li>latLongAcc: Maximum accuracy of latitude longitude coordinates in meters. Minimum 100m
```
mjs: This should be "Prec" not "Acc". You're adjusting precision not accuracy.

test/testdevice.py
```diff
@@ -21,8 +21,10 @@ def has_audio_recording(self):
             self.devicename))
         return self.upload_audio_recording()
 
-    def upload_recording(self):
+    def upload_recording(self, properties = {}):
```
mjs: Mutable objects as keyword arg defaults are dangerous. See https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

It looks like the way you're using it doesn't suffer from the problem but it's best to avoid this kind of thing. Set the default as None and handle that in the function.

test/testdevice.py
```diff
@@ -21,8 +21,10 @@ def has_audio_recording(self):
...
         props = self.get_new_recording_props()
+        for key in properties:
+            props[key] = properties[key]
```
mjs: It looks like you just want to use the `update` dict method.

This would take care of the default argument handling and the property overrides:

```python
props = self.get_new_recording_props()
if properties:
    props.update(properties)
```

PR #143, opened 2018-10-01
remove unused test user

CameronRP:


CameronRP:
I added it before I added the admin API so I could have a user with read permissions for testing but only half finished it. And then mistakenly added it in a pull request.

PR #149, opened 2018-10-26
Fixed group user add and device user add. 

clare:
Now you can use username rather than userid.

Also fixed a couple of problems with the sequelize change.


api/V1/Device.js
```diff
@@ -148,7 +148,7 @@ module.exports = function(app, baseUrl) {
   *
   * @apiUse V1UserAuthorizationHeader
   *
-  * @apiParam {Number} userId ID of the user to delete from the device.
+  * @apiParam {Number} username name of the user to delete from the device.
```
mjs: "Name" not "name"

api/V1/Group.js
```diff
@@ -99,8 +99,8 @@ module.exports = function(app, baseUrl) {
   *
   * @apiUse V1UserAuthorizationHeader
   *
-  * @apiParam {Number} groupId ID of the group.
-  * @apiParam {Number} userId ID of the user to add to the grouop.
+  * @apiParam {Number} group name of the group.
+  * @apiParam {Number} username name of the user to add to the grouop.
```
mjs: capitalisation for these as well

api/V1/Group.js
```diff
@@ -147,8 +147,8 @@ module.exports = function(app, baseUrl) {
   *
   * @apiUse V1UserAuthorizationHeader
   *
-  * @apiParam {Number} groupId ID of the group.
-  * @apiParam {Number} userId ID of the user to remove from the grouop.
+  * @apiParam {Number} group name of the group.
+  * @apiParam {Number} username username of user to remove from the grouop.
```
mjs: and here

PR #150, opened 2018-10-28
added user.isInGroup() and fixed 'interesting' tag query.

CameronRP:


models/User.js
```diff
@@ -208,6 +208,11 @@ module.exports = function(sequelize, DataTypes) {
       });
   };
 
+  User.prototype.isInGroup = async function(id) {
```
clare: MIght be clearer if id, was groupId.

PR #151, opened 2018-10-29
User won't get an error if they try to update to the same email

CameronRP:


models/User.js
```diff
@@ -124,7 +124,7 @@ module.exports = function(sequelize, DataTypes) {
 
   User.freeEmail = async function(email) {
     email = email.toLowerCase();
-    var user = await this.findOne({where: {email: email}});
+    var user = await this.findOne({where: {email: email, id: {[Op.not]: this.id}}});
```
mjs: Can you use `this` here? freeEmail is a class method isn't it?

PR #154, opened 2018-10-31
Fix message when username doesn't exist.  

clare:
(This still isn't great...and could be improved.)


api/middleware.js
```diff
@@ -168,7 +168,7 @@ function getUserByNameOrId(checkFunc) {
   return oneOf([
     getUserByName(checkFunc),
     getUserById(checkFunc),
-  ], "Either 'username' or 'userId' must be specified.");
+  ]);
```
mjs: The message shouldn't have been deleted here

PR #155, opened 2018-11-14
Cptv metadata

clare:
Changes so the metadata can be extracted from the CPTV by the processor.   This is needed as it won't be done when files are uploaded from the sidekick.


api/fileProcessing/index.js
```diff
@@ -158,4 +164,30 @@ module.exports = function(app) {
...
+
+  /**
+   * @api {post} /api/fileProcessing/metadata Add a tag to a recording
+   * @apiName updateMetaData
+   * @apiGroup FileProcessing
+   *
+   * @apiDescription This call updates the metadata for a recording
+   *
+   * @apiParam {Number} recordingId ID of the recording that you want to tag.
+   * @apiparam {JSON} Meta data of recording to ammend.  See /api/V1/recording for more details
```
mjs: This line is missing the param name and amend is spelled wrong. How about:
```
 @apiparam {JSON} metadata Metadata of recording to amend.  See /api/V1/recording for more details.
```

api/fileProcessing/index.js
```diff
@@ -158,4 +164,30 @@ module.exports = function(app) {
       recordingUtil.addTag(request, response);
     })
   );
+
+  /**
+   * @api {post} /api/fileProcessing/metadata Add a tag to a recording
```
mjs: I'm pretty sure this API method doesn't add a tag :)

api/fileProcessing/index.js
```diff
@@ -68,6 +68,7 @@ module.exports = function(app) {
     var jobKey = request.body.jobKey;
     var success = request.body.success;
     var result = request.body.result;
+    var complete = request.body.complete;
```
mjs: This needs to be added to the api docs.

api/fileProcessing/index.js
```diff
@@ -88,6 +89,9 @@ module.exports = function(app) {
         errorMessages.push("'result' field was not a valid JSON.");
       }
     }
+    if (complete == null) {
```
mjs: `if (!complete)` might be safer

test/helper.py
```diff
@@ -110,7 +110,7 @@ def _make_long_name(self, testClass, name):
         testName = type(testClass).__name__
         if testName[:4] == "Test":
             testName = testName[4:]
-        return "{}_{}_{}".format(date.today().strftime('%m%d'), testName, name)
+        return "{}_{}_{}".format(testName, name, date.today().strftime('%m%d'))
```
mjs: better

mjs:
Lint issue needs to be fixed.

PR #156, opened 2018-11-15
Adding indexes to speed up querys when filtering by animals.

CameronRP:


CameronRP:
When I was running the query on my computer for filtering by cat on our database it went from longer than I could wait (laptop was getting quite warm) to less than a second.

PR #159, opened 2018-11-29
 Merge additionalMetadata when updating through file processing API

mjs:
Included is test infrastructure to support testing the file processing API.

### DIFF
--- a/test/test_14_file_processing.py
+++ b/test/test_14_file_processing.py
@@ -1,0 +1,57 @@
+class TestFileProcessing:
+    def test_thermal_video(self, helper, file_processing):
+        user = helper.admin_user()
+
+        # Ensure there's a recording to work with (the file processing
+        # API may return a different one though).
+        helper.given_a_recording(self)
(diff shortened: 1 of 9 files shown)


test/test_14_file_processing.py
```diff
@@ -0,0 +1,57 @@
...
+                "fileMimeType": "application/cheese",
+            },
+            new_object_key=recording["rawFileKey"],
+        )
+        check_recording(
+            user,
+            recording["id"],
+            processingState="FINISHED",
+            fileMimeType="application/cheese",
+            additionalMetadata={"one": "foo", "two": "bar"},  # note: merged
```
clare: Yay new tests! 

Might be better as two tests though.  
1.   That checks the recording state changes and complete works.  
2.  That checks additionalMetadata merges. 

Also may want to check that you can override meta data too.

mjs: ok, I'll split it up

PR #162, opened 2018-12-10
More user friendly error messages

CameronRP:
Mainly just removing the name of the error in the error messages. The name was not useful to the user and was often just "_error"


CameronRP:
@mjs I think so. The main ones for registering and logging in at least.

PR #163, opened 2018-12-18
 Add API to return details of users who can access a device 

mjs:
`GET /api/V1/devices/users`
This can be called by any user who has admin privs for the device.

Also removed duplicated Device.userPermissions method.

### DIFF
--- a/api/V1/Device.js
+++ b/api/V1/Device.js
@@ -90,6 +90,59 @@ module.exports = function(app, baseUrl) {
   );
 
   /**
+  * @api {get} /api/v1/devices/users Get all users who can access a device.
+  * @apiName GetDeviceUsers
+  * @apiGroup Device
+  * @apiDescription Returns all users that have access to the device
+  * through both group membership and direct assignment.
+  *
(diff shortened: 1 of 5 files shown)


api/V1/Device.js
```diff
@@ -89,6 +89,59 @@ module.exports = function(app, baseUrl) {
...
+  * @apiSuccess {JSON} rows Array of users who have access to the
+  * device. Each element includes `id` (user id), `username`, `email`,
+  * `relation` (either `group` or `device`) and `admin` (boolean).
+  * @apiUse V1ResponseError
+  */
+  app.get(
+    apiUrl + '/users',
+    [
+      middleware.authenticateUser,
+      middleware.getDeviceById(body),
```
CameronRP: You will want to use `query` here instead of `body` for a GET request.
The middleware looks to load the device object into the body though.

mjs: Good catch! The test code is making the request incorrectly too so I didn't notice.

Serves me right for copying and pasting.

PR #166, opened 2018-12-19
Tidy up Events api

mjs:
* Tighten up handling of get events query params
  - Automatic int conversion
  - Fixed bug when setting default for `limit`
  - Cleaner default handling
* Upgrade express-validator to address various sanitisation bugs
* Rationalise query test helpers
* Replace Events query "where" param with separate query params. This is safer and helps to avoid lock-in to sequelizejs.

### DIFF
--- a/api/V1/Event.js
+++ b/api/V1/Event.js
@@ -21,6 +21,7 @@ const responseUtil = require('./responseUtil');
 const middleware   = require('../middleware');
 const { body, query, oneOf } = require('express-validator/check');
 
+
 module.exports = function(app, baseUrl) {
   var apiUrl = baseUrl + '/events';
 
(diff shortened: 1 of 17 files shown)


api/V1/Event.js
```diff
@@ -111,49 +112,49 @@ module.exports = function(app, baseUrl) {
   );
 
   /**
-   * @api {get} /api/v1/events Query events recorded
+   * @api {get} /api/v1/events Query recorded events
    * @apiName QueryEvents
    * @apiGroup Events
    *
    * @apiUse V1UserAuthorizationHeader
+   * @apiParam {Datetime} startTime Return only events after this time (optional)
```
CameronRP: You can make them optional with apidoc. I think you just need to put the field name in square brackets. So
 `@apiParam {Datetime} [startTime] Return only events after this time`
http://apidocjs.com/#param-api

mjs: Ah! Nice. I will fix.

PR #167, opened 2018-12-28
Calculate success flag

### LINKED ISSUE #165: responseUtil.send should not access `status`
The `send` function in `api/V1/responseUtil.js` should automatically generate the success boolean based on the HTTP status code instead of requiring that it is passed in.

All call sites need to be updated to match.

agronauts:
Fixes #165

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,9 @@ Once the container is running, you can start a bash session inside
 the container with `docker exec -it cacophony-api bash`.  You can then use
 psql to query the database `sudo -i -u postgres psql cacophonytest`
 
+Note that the source code is mounted on to the docker image, so any changes 
+causes the server to reload.
+
 ## Running the tests
 
 The Cacophony API server has a comprehensive function test suite. This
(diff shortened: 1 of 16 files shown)


run
```diff
@@ -24,6 +24,10 @@ parser.add_argument(
 )
 args = parser.parse_args()
 
+if not os.path.exists("node_modules"):
+    print("Node modules have not been installed yet, doing so now")
+    subprocess.call(["npm", "install"])
```
mjs: nice - thanks

README.md
```diff
@@ -29,6 +29,9 @@ Once the container is running, you can start a bash session inside
 the container with `docker exec -it cacophony-api bash`.  You can then use
 psql to query the database `sudo -i -u postgres psql cacophonytest`
 
+Note that the source code is mounted on to the docker image, so any changes 
+causes the server to reload.
```
mjs: Thanks for adding this. Can you please extend this further to mention `./run --isolate` which runs the server independently of the source tree (including `node_modules`)?

PR #177, opened 2019-03-11
Algorithm snapshot

clare:
Generalised the EventDetails idea so that it can be used to store settings (which do not change very often).   Renamed the table to DetailSnapshots and used this to store the algorithm settings for the tracks.


.travis.yml
```diff
@@ -21,3 +21,4 @@ script:
  - test/check-style
  - ./wait-for-api || { docker ps; docker logs cacophony-api; exit 1; }
  - cd test && pytest
+ - docker exec -it cacophony-api /app-isolated/node_modules/.bin/sequelize db:migrate:undo --config /app-isolated/config/app_test_default.js --migrations-path /app-isolated/migrations
```
mjs: This appears to undo the latest DB migration. Is that particularly useful without a more detail check afterwards?

api/fileProcessing/index.js
```diff
@@ -194,7 +194,8 @@ module.exports = function(app) {
    * @apiGroup FileProcessing
    *
    * @apiParam {JSON} data Data which defines the track (type specific).
-   * @apiParam {Number} algorithm Tracking algorithm version number.
+   * @apiParam {Number} AlgorithmId Database Id of the Tracking algorithm details retrieved from
+   * <a href="#api-fileprocessing-algorithm">Algorithm</a> request
```
mjs: You can link to other sections like this: `(#Group:Name)`. So in this case `(#FileProcessing:Algorithm)`

api/fileProcessing/index.js
```diff
@@ -323,4 +322,31 @@ module.exports = function(app) {
...
+    apiUrl + '/algorithm',
+    [
+      middleware.parseJSON('algorithm', body),
+    ],
+    middleware.requestWrapper(async (request, response) => {
+      var algorithm = await models.DetailSnapshot.getOrCreateMatching("algorithm", request.body.algorithm);
+
+      responseUtil.send(response, {
+        statusCode: 200,
+        messages: ["Algorithm key retreived."],
```
mjs: retrieved

PR #178, opened 2019-03-20
Fix track tagging so that user is recorded and username

clare:


models/Recording.js
```diff
@@ -317,6 +317,18 @@ module.exports = function(sequelize, DataTypes) {
     }
   };
 
+  /* eslint-disable indent */
```
mjs: why disable this here?

clare: Just disables the indenting lint messages.  I've indented it so that you can see which requirement is on which table.

PR #183, opened 2019-03-25
Ignore old tracks

gferraro:
Added 2 new API endpoints to reprocess a recording and multiple recordings:
  * @api {get} /api/v1/recordings/reprocess/:id
  * @api {post} /api/v1/recordings/reprocess/multiple


api/V1/Recording.js
```diff
@@ -477,6 +478,54 @@ module.exports = (app, baseUrl) => {
...
+      middleware.authenticateUser,
+      param('id').isInt(),
+    ],
+    middleware.requestWrapper(async (request, response) => {
+      return await recordingUtil.reprocess(request,response)
+    })
+  );
+
+  /**
+  * @api {post} /api/v1/recordings/reprocess/multiple marks recordings for reprocessing and archives tracks
```
mjs: Can this just be a POST to `/api/v1/recordings/reprocess`? I don't think the "multiple" part is necessary.

api/V1/recordingUtil.js
```diff
@@ -187,8 +187,72 @@ async function addTag(request, response) {
   });
 }
 
+// reprocessAll expects request.body.recordings to be a list of recording_ids
+// will mark each recording to be reprocessed
+async function reprocessAll(request, response) {
+  recordings =  request.body.recordings
+  var responseMessage = {
```
mjs: it's a good habit to use `const` to avoid unintended reassignment (`const` doesn't mean what it might in other languages)

api/V1/recordingUtil.js
```diff
@@ -187,8 +187,72 @@ async function addTag(request, response) {
...
+    statusCode: 200,
+    messages: [],
+  }
+  for (var i = 0; i < recordings.length; i++){
+    resp = await reprocessRecording(request.user, recordings[i])
+    if(resp.statusCode != 200){
+      responseMessage.statusCode = resp.statusCode
+      responseMessage.push(resp.messages[0])
```
mjs: is this supposed to be `responseMessage.messages.push`?

api/V1/recordingUtil.js
```diff
@@ -187,8 +187,72 @@ async function addTag(request, response) {
   });
 }
 
+// reprocessAll expects request.body.recordings to be a list of recording_ids
+// will mark each recording to be reprocessed
+async function reprocessAll(request, response) {
+  recordings =  request.body.recordings
+  var responseMessage = {
+    statusCode: 200,
+    messages: [],
```
mjs: how about also returned the ids of the ones that succeeded?

api/V1/recordingUtil.js
```diff
@@ -187,8 +187,72 @@ async function addTag(request, response) {
...
+      statusCode: 400,
+      messages: ["No such recording or access denied " + recording_id],
+    }
+  }
+
+  await models.Track.update({
+      archivedAt:Date.now()
+    },
+    {
+      where: { RecordingId:recording.id,archivedAt:null }
```
mjs: space after comma

api/V1/recordingUtil.js
```diff
@@ -187,8 +187,72 @@ async function addTag(request, response) {
...
+    }
+  }
+
+  responseUtil.send(response, responseMessage);
+  return
+}
+
+// reprocessRecording marks supplied recording_id for reprocessing, 
+// under supplied user privilages
+async function reprocessRecording(user,recording_id){
```
mjs: This guts of this should probably be a method on a Recording model object. Then we can retrieve the recording and call `rec.reprocess(user)` .  Of course, that shouldn't be returning HTTP status codes.

api/V1/Recording.js
gferraro: Maybe i'm missing something but if I don't put a word at the end it gets confused with path:
   * @api {post} /api/v1/recordings/:devicename 
I suppose a case is that devicename could be "reprocess"

api/V1/recordingUtil.js
```diff
@@ -220,32 +250,23 @@ async function reprocessRecording(user,recording_id){
     return {
       statusCode: 400,
       messages: ["No such recording or access denied " + recording_id],
-    }
+      id: recording_id
```
mjs: Can this be recordingId to be consistent with naming elsewhere?

api/V1/Recording.js
mjs: Bummer. What if you put a trailing "/" after "reprocess" in the URL spec?

Failing that, just leave it how it is.

api/V1/recordingUtil.js
```diff
@@ -187,8 +187,93 @@ async function addTag(request, response) {
...
+    statusCode: 200,
+    messages: [],
+    reprocessed: [],
+    fail: [],
+  };
+
+  var status = 0;
+  for (var i = 0; i < recordings.length; i++){
+    var resp = await reprocessRecording(request.user, recordings[i]);
+    if(resp.statusCode != 200){
```
mjs: i'm surprised that eslint isn't complaining about the lack of spaces here :)

api/V1/recordingUtil.js
```diff
@@ -187,8 +187,93 @@ async function addTag(request, response) {
...
+      status = status | statusCode.Fail;
+      responseMessage.messages.push(resp.messages[0]);
+      responseMessage.statusCode = resp.statusCode;
+      responseMessage.fail.push(resp.id);
+    }else{
```
mjs: and here

Issue #19, opened 2017-10-04
Registration requests for already existing users should fail with a 4xx error 


mjs:
Currently they fail with a 5xx.

jamie-sherriff:
@mjs Was scanning over good first issues and i had a quick look at this at this fails with a 422 error now? Is this the expected behavior?

`info: POST /api/v1/Users
warn: Validation error[422]: Username in use; Email in use.
info: POST /api/v1/Users 422 9ms`

If so would you like me to write some tests to cover this condition?

mjs:
@jamie-sherriff So the issue is already fixed but there's no test coverage? If so, a test for this would great. Thanks!

mjs:
Fixed!

PR #195, opened 2019-04-09
Add a test for Issue #19 to show working as expected

jamie-sherriff:
Went a bit further and parsed the JSON in the error message, any thoughts on breaking it out into another test or just leave it out?

### DIFF
--- a/test/test_user.py
+++ b/test/test_user.py
@@ -1,3 +1,5 @@
+import json
+
 class TestUser:
     def test_can_create_new_user(self, helper):
         print("If a new user Bob signs up", end="")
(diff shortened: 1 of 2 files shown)


test/test_user.py
```diff
@@ -27,3 +30,24 @@ def test_can_create_new_user(self, helper):
...
+            try:
+                helper.given_new_fixed_user(
+                    "DuplicateBob"
+                )
+            except OSError as error:
+                print("DuplicateBob should not get created multiple times and returns a 422 error message")
+                error_string = str(error)
+                assert "request failed (422)" in error_string
+                print("There should be a JSON object in the error message that can be parsed")
+                json_object = error_string[(error_string.find(":")+1):]
```
mjs: Something like `json_object = error_string.split(":", 1)[1:]`  would be a little cleaner

test/test_user.py
```diff
@@ -27,3 +30,24 @@ def test_can_create_new_user(self, helper):
...
+                parsed_json = json.loads(json_object)
+                assert parsed_json['errorType'] == "validation"
+                assert (
+                        "'Username in use" in parsed_json['message']
+                        or "Email in use" in parsed_json['message']
+                )
+                print("DuplicateBob already exists")
```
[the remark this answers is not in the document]
jamie-sherriff: Ah thanks! i hadn't thought of that, have applied the change

test/test_user.py
jamie-sherriff: Have made the change

PR #196, opened 2019-04-09
All tags

clare:
So that we can release the new track tag gui, we need to be able to see what tags are on each recording.  

Retrieving recordings with tags now retrieves recordings tagged at track level and recording level.   Also improved options now allow you to retrieve particular types of tags (eg ai vs human) and different whats (animals etc)

### DIFF
--- a/models/Recording.js
+++ b/models/Recording.js
@@ -131,7 +131,10 @@ module.exports = function(sequelize, DataTypes) {
       order: order,
       include: [
         { model: models.Group, where: {}, attributes: ["groupname"] },
-        { model: models.Tag, where: {}, attributes: ["animal", "automatic", "event", "taggerId"], required: false },
+        { model: models.Tag,
+          where: {},
+          attributes: ["animal", "automatic", "event", "taggerId"],
+          required: false },
         { model: models.Track,
           where:{
             archivedAt: null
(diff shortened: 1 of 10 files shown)


models/Recording.js
```diff
@@ -130,13 +130,21 @@ module.exports = function(sequelize, DataTypes) {
       },
       order: order,
       include: [
-        { model: models.Group },
-        { model: models.Tag, where: makeTagsWhere(tags), include: [{association: 'tagger', attributes: ['username']}] },
-        { model: models.Device, where: {}, attributes: ["devicename", "id"] },
+        { model: models.Group, attributes: ["groupname"] },
+        { model: models.Tag,where: makeTagsWhere(tags), attributes: ["animal", "automatic", "event", "taggerId"] },
```
mjs: space after comma

test/conftest.py
```diff
@@ -45,5 +45,5 @@ def pytest_runtest_makereport(item, call):
             )
         except FileNotFoundError:
             return
-        if proc.returncode == 0:
-            rep.sections.append(("Recent API server logs", proc.stdout))
+        # if proc.returncode == 0:
+        #     rep.sections.append(("Recent API server logs", proc.stdout))
```
mjs: Can this be put back please? 

Have it check an environment variable (via Python's `os.environ`) to decide whether to include this section. I'm ok with the default being off. We can turn it on for travis runs.

test/recording.py
```diff
@@ -1,8 +1,10 @@
 class Recording:
-    def __init__(self, id_, props, content_filename):
+    def __init__(self, id_, props, content_filename, recording_name=''):
         self.id_ = id_
         self.props = props
         self.content = slurp(content_filename)
+        # this name is just for test purposes
```
mjs: what do you mean by test purposes? can you make the comment more descriptive?

PR #197, opened 2019-04-11
Modify run file to support without using sudo

jamie-sherriff:
I added my user to the docker user group so i don't need to prefix docker commands with sudo and password. Have left the default run using sudo.

### DIFF
--- a/run
+++ b/run
@@ -3,6 +3,10 @@
 import argparse
 import os
 import subprocess
+import getpass
+
+if os.name == "posix": # Assume linux/MacOs variants
+    import grp # pylint: disable=import-error
 
 IMAGE_NAME = "cacophony-api"
 CONTAINER_NAME = "cacophony-api"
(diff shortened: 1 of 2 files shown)


run
```diff
@@ -22,21 +22,34 @@ parser.add_argument(
...
 args = parser.parse_args()
-
 if not os.path.exists("node_modules"):
     print("Node modules have not been installed yet, doing so now")
     subprocess.call(["npm", "install"])
 
 print("Stopping {} container (if running)".format(CONTAINER_NAME))
+remove_container_cmd = ["sudo", "docker", "rm", "--force", CONTAINER_NAME]
+if args.no_sudo:
+    remove_container_cmd.remove("sudo")
```
mjs: Instead of removing an element how about only adding it if required? That's probably cleaner.

Something like:
```python
cmd = []
if not args.no_sudo:
    cmd.append("sudo")
cmd = ["docker", "rm", "--force", CONTAINER_NAME]
...
```

It would be even cleaner if you reversed the sense of the option: `args.sudo` instead of `args.no_sudo`. Code generally tends to work out cleaner when flags refer to the positive case.

run
```diff
@@ -22,21 +22,34 @@ parser.add_argument(
...
+    remove_container_cmd,
     stdout=subprocess.DEVNULL,
     stderr=subprocess.STDOUT,
 )
 
 print("Building container")
-subprocess.check_call(["sudo", "docker", "build", ".", "-t", IMAGE_NAME])
+build_container_cmd = ["sudo", "docker", "build", ".", "-t", IMAGE_NAME]
+if args.no_sudo:
+    build_container_cmd.remove("sudo")
```
mjs: Given that this pattern is repeated you could clean things up by having a function which optionally adds sudo. 

If you wanted to go even further and you could define a function dynamically like this depending on the option:

```python
parser.add_argument(
    "--no-sudo",
    default=False,
    action="store_true",
    help="Run commands without sudo",
)
args = parser.parse_args()


if args.no_sudo:
   add_sudo = lambda cmd: cmd
else:
   add_sudo = lambda cmd: ["sudo"] + cmd
```

Then later the code that constructs the commands just end up looking like...
```python
build_container_cmd = add_sudo(["docker", "build", ".", "-t", IMAGE_NAME])
```

No ifs required, less lines of code, less chance for errors.

run
jamie-sherriff: Agreed! i was just erring on the side of caution to default to using sudo first.

run
jamie-sherriff: Have implemented

run
```diff
@@ -22,25 +26,58 @@ parser.add_argument(
...
+    default=False,
+    action="store_true",
+    help="Run commands with sudo",
+)
 args = parser.parse_args()
 
+use_sudo = args.sudo
+if not args.sudo:
+    current_user = getpass.getuser()
+    user_in_docker_group = current_user in grp.getgrnam("docker").gr_mem
```
mjs: This feels like it should be in its own function (`is_user_in_group()` or similar)

run
```diff
@@ -22,25 +26,58 @@ parser.add_argument(
...
+    if user_in_docker_group:
+        print("{} is in docker group".format(current_user))
```
mjs: there's probably no need to print this every time.

run
```diff
@@ -22,25 +26,58 @@ parser.add_argument(
...
+    #Run docker ps without sudo to check if it can be used without
+    if not user_in_docker_group:
+        return_code = subprocess.call(
+            ["docker", "ps"],
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            timeout=30
+        )
+        print("docker ps returned: {}".format(return_code))
```
mjs: nor this

run
```diff
@@ -22,25 +26,58 @@ parser.add_argument(
...
+        if return_code != 0:
+            use_sudo = True
```
mjs: checking `docker ps` feels like it should be in it's own function too

run
```diff
@@ -3,6 +3,10 @@
 import argparse
 import os
 import subprocess
+import getpass
+
+if os.name == "posix": # Assume linux/MacOs variants
```
mjs: is this guard really that useful? `sudo` doesn't work on Window either and without importing `grp` later code will crash anyway

run
jamie-sherriff: done

run
jamie-sherriff: No problem

run
jamie-sherriff: Removed as well, have left in the one print with sudo status though

run
jamie-sherriff: Done

run
jamie-sherriff: Yeah good point

PR #198, opened 2019-05-02
Fixes #128. Refactored auth handling to use standard Express middleware and return 401/403

### LINKED ISSUE #128: Server returns 422 status when authentication token is invalid
When the JWT is missing or invalid the API should return 401 not 422./

jackodsteel:
Refactored auth handling to use standard Express middleware instead of validators, allowing us to directly return 401/403 status codes when relevant. Fixes #128 

Also refactored a lot of authorization handling in methods to throw a newly added AuthorizationError. Added a global error handler to the requestWrapper, allowing the Auth errors to bubble up directly and then be correctly handled.

Added custom Exceptions to the python test suite as well to more easily and specifically handle auth rejections.
Reworked all tests to expect a 403 when they should be forbidden rather than a 422.
Added a helper method to raise the correct exception if there is an error.

### DIFF
--- a/api/V1/Admin.js
+++ b/api/V1/Admin.js
@@ -17,6 +17,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 const middleware   = require('../middleware');
+const auth         = require('../auth');
 const models       = require('../../models');
 const responseUtil = require('./responseUtil');
 const { param, body } = require('express-validator/check');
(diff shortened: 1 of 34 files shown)


jackodsteel:
Happy for plenty of feedback/review on this BTW, there's a decent amount changed and although the tests pass it's definitely possible some stuff is incorrect.

It would probably be good to add tests for authentication as well at some point.

clare:
There are a just a few checkstyle errors.   You can hopefully see them https://travis-ci.com/TheCacophonyProject/cacophony-api/builds/110685975. 

Or run >npm run lint
on the project to see them.   

Big thanks!

jackodsteel:
Awesome :) @clare Should be all good to go now so not sure the exact process with the smoke testing stuff and if you want to merge to master while that's still ongoing (seems to be working as expected), but I'm happy for you to merge when ready.

clare:
Thanks Jack!   Awesome that it's all finished :)

Issue #56, opened 2018-02-02
Processing jobs being given out multiple times for the same recording


mjs:
It's looking like the file processing API is sometimes giving out the same recording to multiple processing instances. This is likely to be because transactions aren't being used properly when setting the processing state for a recording.

jackodsteel:
Fixed in #201

PR #201, opened 2019-05-02
Added transaction to file processing get API

gferraro:
fixes issues #56

### DIFF
--- a/test/test_file_processing.py
+++ b/test/test_file_processing.py
@@ -3,9 +3,33 @@ from datetime import datetime, timezone
 from .recording import Recording
 from .track import Track
 from .track import TrackTag
+from multiprocessing import Pool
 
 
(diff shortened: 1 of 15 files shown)


test/test_file_processing.py
```diff
@@ -14,7 +38,7 @@ def test_thermal_video(self, helper, file_processing):
         helper.given_a_recording(self)
 
         # Get a recording to process.
-        recording = file_processing.get("thermalRaw", "getMetadata")
+        file_processing.get("thermalRaw", "getMetadata")
         assert recording["processingState"] == "getMetadata"
```
CameronRP: Doesn't look like recording is defined anymore?

gferraro: oh yep sorted now

PR #202, opened 2019-05-08
Fixed tag deletion api

gferraro:
Fixes taiga-570
Checks user permissions to recording, to see if can delete tags.
Also added test cases.


models/Tag.js
```diff
@@ -75,17 +76,22 @@ module.exports = function(sequelize, DataTypes) {
...
+      user,
+      tag.RecordingId,
+      Recording.Perms.TAG,
+    );
+
+    if(recording == null){
+      return false;
     }
-    else {return false;}
+    return tag.destroy();
```
mjs: shouldn't this be:
```js
await tag.destroy();
return true;
```
?

Otherwise sometimes it returns false and sometimes it returns a promise which returns who knows what.

test/test_tagging.py
```diff
@@ -1,15 +1,37 @@
...
+        untagged = device.has_recording()
+
+        tag = untagged.is_tagged_as("possum").by(lucy)
+        lucy.delete_recording_tag(tag["tagId"])
+
+        tag = untagged.is_tagged_as("cat").by(lucy)
+        admin.delete_recording_tag(tag["tagId"])
+
+        tag = untagged.is_tagged_as("cat").by(lucy)
+        sophie.cannot_delete_recording_tag(tag["tagId"])
```
mjs: thank you for adding tests!

models/Tag.js
gferraro: The promise should return 1 if successful and 0 if unsuccessful, I figured because the function is async a promise is an expected result. Seems strange to have all these async functions with awaits in them?

mjs: It's totally fine for an async function to return a value. You can only use `await` in an async function. When an async function is called (often with await) it means the JS runtime is allowed to run other things during the function call (e.g. when it calls await on other functions).

An async function can also return promises but then the return value ends up being the return value of that promise. In this case `tag.destroy()` returns a promise which returns the number of deleted rows. It's messy to sometimes return `false`, sometimes `1` and sometimes `0`.

Also, it should be considered an error if a tag was to be deleted that was already deleted. This is success as the intent of the operation has happened.

gferraro: Turns out I was reading the wrong doc for destroy anyway. It just returns a promise.

PR #219, opened 2019-06-12
parse success as a boolean

gferraro:
fix bug where adding tags failed due to no user from file api

### DIFF
--- a/api/middleware.js
+++ b/api/middleware.js
@@ -170,7 +170,9 @@ const parseArray = function(field, checkFunc) {
   });
 };
 
-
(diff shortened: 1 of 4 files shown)


api/middleware.js
```diff
@@ -170,7 +170,9 @@ const parseArray = function(field, checkFunc) {
   });
 };
 
-
+const parseBool = function(value) {
+  return (value == "true") || (value == "True");
```
mjs: maybe just lowercase the input first so you only have to check one value and it works for any case?

PR #221, opened 2019-06-14
changed device index constraint to be on groupid + devicename

gferraro:
This will break a few things
@api {post} /authenticate_device/ 
now requires groupname in postdata

@api {post} /api/v1/recordings/:devicename 
becomes 
@api {post} /api/v1/recordings/device/:devicename/group/:groupname?

@api {get} api/v1/schedules/:devicename
becomes
@api {get} api/v1/schedules/:deviceId

@api {get} recording/reprocess
becomes
@api {get} /api/v1/reprocess/:id
@api {post} /api/v1/reprocess for multiple

devicenames now must contain atleast one alphabetic character.
devicenames still must be unique.
devices will be found as per https://docs.google.com/document/d/1nRhlNpwv6k0vm-HkzgF2xzB4R670tzP4BfqRm-316xU/edit


api/V1/Recording.js
```diff
@@ -84,11 +84,41 @@ module.exports = (app, baseUrl) => {
    * @apiSuccess {Number} recordingId ID of the recording.
    * @apiuse V1ResponseError
    */
+
   app.post(
-    apiUrl + "/:devicename",
+    apiUrl + "/device/:devicename/group/:groupname?",
```
mjs: what does the `?` on the end mean?

api/V1/Recording.js
```diff
@@ -84,11 +84,41 @@ module.exports = (app, baseUrl) => {
...
+    middleware.requestWrapper(recordingUtil.makeUploadHandler())
+  );
+
+  /**
+   * @api {post} /api/v1/recordings/device/:deviceId Add a new recording on behalf of device
+   * @apiName PostRecordingOnBehalf
+   * @apiGroup Recordings
+   * @apiDescription Called by a user to upload raw thermal video on behalf of a device.
+   * The user must have permission to view videos from the device or the call will return an
+   * error.  It currently supports raw thermal video but will eventually support all
```
mjs: Please remove the bit about this only working for thermal video. Same for the other POST API that takes group name

api/V1/Reprocess.js
```diff
@@ -0,0 +1,67 @@
+/*
+cacophony-api: The Cacophony Project API server
+Copyright (C) 2018  The Cacophony Project
```
mjs: 2019

test/test_device_group_names.py
```diff
@@ -0,0 +1,90 @@
...
+
+    #     description = f" there can also be a device called '{c_terminator.devicename}' in this group"
+    #     p_terminator = helper.given_new_device(
+    #         None, c_terminator.devicename, pats_group, description=description
+    #     )
+
+    #     devices = pat.get_devices_as_ids()
+    #     assert devices == [p_terminator.get_id()]
+    #     devices = clare.get_devices_as_ids()
+    #     assert devices == [c_terminator.get_id()]
```
mjs: are these all supposed to be commented out?

api/V1/Recording.js
gferraro: optional, so that it works with old bird-monitor releases

test/test_device_group_names.py
gferraro: ye they are based of 2 devicenames being the same. I moved the test about deivcenames having to have a alphabetic character

test/test_device_group_names.py
```diff
@@ -20,8 +20,16 @@ def test_authentication(self, helper):
...
+            c_terminator = helper.given_new_device(
+                None,
+                "1234",
+                clares_group,
+                description="and devicename 1234 fails, for it must containt alpha char",
+            )
+
     # use this when we make group + device unique
-    # def test_authentication(self, helper):
+    # def test_unique_authentication(self, helper):
```
mjs: Rather than comment out the tests, use pytest's helpers for skipping tests. These take a reason which will be displayed when the test is skipped.

https://docs.pytest.org/en/latest/skipping.html

This will make it clearer that there are some tests that are intentionally being skipped (makes them harder to forget).

PR #233, opened 2019-06-21
Initial work to convert run script to docker-compose and allow easier Windows development, improve Docker usage.

jackodsteel:
The current Docker setup is suboptimal and doesn't work on Windows for development.

This is my initial work to refactor the `Dockerfile` to be more inline with the best practices as well as improving development experience.

I'm trying to refactor away the `run` bash file, as it's platform specific and can hopefully be replaced with `docker-compose`.

## Dockerfile changes
I've done two main things to the Dockerfile. Firstly, I've tried to combine similar steps to reduce the layers as is the best practice pattern with Dockerfiles, and also reworked the `apt-get` part, as it's also best practice to combine these, and to run the `apt-get update` in the same `RUN` command as the `install`, then cleanup afterwards. It just helps to have cleaner images.

The main change I've made though, is to copy the source in to the image in two steps. Firstly, copy the `package.json` and run `npm install`, then copy the source code. This is a common best practice for dockerizing Node projects. Doing it in two steps means that code changes don't need to rebuild the node modules which takes most of the time.
By copying this stuff in rather than forcing it to be mounted, it allows more stable tests, but also allows dev work on Windows (as mounting it in from Windows doesn't work).

## Adding docker-compose
I'm trying to get a nice docker-compose setup going which emulates what the `run` script is doing. The idea is that the `docker-compose.yml` should be the prod/non live update version, but you can also run the `docker-compose.live-dev.yml` to mount the source code in as you're already doing, which will override the source code that's been copied in to the actual image.

# Testing
I don't use the live dev or advanced features, so if someone that does could try and see if their workflow works using the docker-compose setup that'd be awesome.
You'll just need to [install docker-compose](https://docs.docker.com/compose/install/) if not already, then try using the following command instead of the `run` script:
```
docker-compose -f docker-compose.yml -f docker-compose.live-dev.yml up --build --force-recreate --detach
```
Hopefully, it should build the docker image and then run the image with the code mounted, and use nodemon to live reload. Let me know if it works as expected, or if any other workflows don't work as expected.

## Next steps
Once it's working I'll update the readme etc.
Once we get this part sorted, it'd possibly be nice to break the Minio/Postgres/api parts in to separate Docker containers and use compose to build them together, but may not be worth the effort.


mjs:
Thanks @jackodsteel for doing this - there's lots of great stuff in here. Sorry it's taken me so long to get to reviewing this.

Some feedback:
* I'm happy to replace `run` with docker-compose, especially if it helps make things easier on Windows
* I can confirm that live reload still works 
* It's worth noting that the application code and node_modules was previously being kept isolated for CI test runs (`run --isolate`). In th
* To avoid the need for people to use long docker-compose command lines a helper script(s) for (re)starting the containers

mjs:
One more little thing: can you set up the container name to be a little less clumsy. "cacophony-api_cacophony-api_1" is a bit much. How about "cacophony-api_app" or "cacophony-api_server"

jackodsteel:
> Sorry it's taken me so long to get to reviewing this.

Heh, I think my reply took a little longer so I guess we'll cut you some slack this time 😉 

I have made it past exams and things so thankfully I've been able to make it back to a few of the things I had planned here!

> * It's worth noting that the application code and node_modules _was_ previously being isolated when `run --isolate` was used. Here's the code was copied in and `npm install` run inside the container after the container had started up. This was mainly used for CI test runs.
> * Having the Dockerfile unconditionally copy in the application code and run `npm install` even when live reload is going to be used makes the image build somewhat slower. I can live with it but can you think of a way to separating this out that's hopefully not too complicated?

Yeah, the two of these are kind of combined and are a good point. We probably could find a way to avoid loading the node_modules and doing the install. However, in what I've found it's generally best practice to copy them in always. Obviously with nodemon it's redundant, but the startup times are pretty short so IMO it's not a huge issue. If it's a turnoff for you guys though then I'm happy to find a solution.
Also remember that the `npm install` step is only going to get run if the `package.json` changes, which won't happen that often. The normal copy is almost instant (< 1 second).

> * To avoid the need for people to use long docker-compose command lines a helper script(s) for (re)starting and stopping the containers would be nice. These would replace the current `run` script. These could be cross-platform Python scripts or there could be separate Unix shell scripts and Windows batch/Powershell scripts.

I've kept the existing `run` script as it does work well. Since it's already Python it should mostly be usably in Windows. I'll have to try it out as I think some of the libraries are OS dependent at the moment, but I'm sure we can fix that.
I've modded it to work with the new compose setup and also added a `--isolate-node-modules` which mounts the app but not the node_modules, which is the best dev setup for Mac/Windows since their node_modules aren't compatable with Linux, but means they can still get the nodemon code reload for everything except dependency changes.

> * If we're going to use docker-compose anyway, I'd like to see Minio and PostgreSQL be separated out. Hopefully, standard Docker images could be used for these reducing the overall build/startup time (especially in TravisCI). This certainly doesn't have to happen in this PR of course!

This is done! It makes the actual app container much simpler which is a good change.

> One more little thing: can you set up the container name to be a little less awkward? 

Fixed!

> So overall: let's get this finalised. There's just a few things to sort out.

No rush on another review, but let me know when you take a look and if there's more to change!

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    container_name: cacophony-api_server
+    build: .
+    environment:
+      NODE_BIN: node
+    ports:
+      - 1080:1080
+      - 2008:2008
+    depends_on:
+      db:
+        condition: service_healthy
```
hardbyte: Note the `condition` is removed in newer versions of docker-compose.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+  s3:
+    container_name: cacophony-api_s3
+    image: minio/minio:latest
+    entrypoint: sh
+    command: -c 'mkdir -p /buckets/cacophony && /usr/bin/minio server --address :9001 /buckets'
+    environment:
+      MINIO_ACCESS_KEY: minio
+      MINIO_SECRET_KEY: miniostorage
+    ports:
+      - 9001:9001
```
hardbyte: Did you consider mounting volumes for the database and object store?

hardbyte: Oh and for `minio` **if** you don't need to create the `/buckets/cacophony` directory, you could simplify by dropping the entrypoint and having the super simple: `command: server /buckets`

clare:
Hi @jackodsteel @mjs I've been having a look at this pull request.   There is a lot of super clever stuff in it but unfortunately it's got a bit out-of-date with the current master branch.   

I also have a couple of concerns:
  1.   I'm not sure if it is ideal to have everything start up in its own docker container.  I don't have much ram on my computer (not linux) and this is slowing it down massively. 
  2.   The new system while very clever is very complex.   There are multiple build files and multiple compose files.   This means everyone is running a slightly different version and there is more to keep working.   I was wondering if it is better to just use docker properly (so the build time is short) and then everyone should use the same settings (isolated node-modules, continuous updates)

Cheers clear.

clare:
Hi team, 

Thanks so much for this pull request.   I have now used this code to create a slightly different pull request https://github.com/TheCacophonyProject/cacophony-api/pull/322.   The main difference is that everyone should run docker using the same commands so that we don't have problem with some variations working and others not.  

Thanks so much @jackodsteel and @hardbyte 

Regards
Clare

PR #236, opened 2019-07-04
Make object pruning safer and run pruning daily

mjs:
* Use a type prefix for the object keys: This allows prune-objects to more safely remove unused keys by only operating on these prefixes. These prefixes could also help make management of the object store easier.
* prune-objects now only acts on specific object key prefixes 
* Run prune-objects daily via cron

### DIFF
--- a/prune-objects.js
+++ b/prune-objects.js
@@ -4,6 +4,13 @@ const { Client } = require("pg");
 const config = require("./config");
 const modelsUtil = require("./models/util/util");
 
+// Define the types of object keys that will be considered for pruning.
+const keyTypes = Object.freeze([
+  { prefix: "f", table: "Files", column: "fileKey" },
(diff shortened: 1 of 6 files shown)


prune-objects.js
```diff
@@ -83,20 +99,13 @@ async function pgConnect() {
...
-async function allDBKeys(client) {
+async function loadDBKeys(client, table, column) {
   const keys = new Set();
-  const res = await client.query(`
-        select "fileKey" as fk, "rawFileKey" as rk from "Recordings"
-        union
-        select "fileKey" as fk, NULL as rk from "Files"
-  `);
+  const res = await client.query(
+    `select "${column}" as k from "${table}" where 1 is not NULL`
```
gferraro: why the where clause?

mjs: `1` refers to the first (and only) column. This just removes the rows where the field is not set.

PR #240, opened 2019-07-10
Change requirement for group and device names

CameronRP:
Can no longer have " " or "_" in names


mjs:
Waiting for feedback on this

PR #249, opened 2019-07-31
Device rename endpoint

CameronRP:
Devices can change there name and group

### DIFF
--- a/api/V1/Device.js
+++ b/api/V1/Device.js
@@ -263,42 +263,14 @@ module.exports = function(app, baseUrl) {
     ],
     middleware.requestWrapper(async function(request, response) {
 
-      const newName = request.body.newName;
-      const newGroup = request.body.group;
+      await request.device.rename(request.body.newName, request.body.group);
 
-      const conflictingDevice = await models.Device.findOne({
-        where: {
-          devicename: newName,
-          GroupId: newGroup.id,
-        }
-      });
-
-      const successResponse = {
+      return responseUtil.send(response, {
         statusCode: 200,
         devicename: request.body.newName,
-        groupname: newGroup.groupname,
-        messages: ["name and gruop set"],
-      };
-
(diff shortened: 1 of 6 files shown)


api/V1/Device.js
```diff
@@ -239,4 +239,67 @@ module.exports = function(app, baseUrl) {
...
+          devicename: newName,
+          GroupId: newGroup.id,
+        }
+      });
+
+      const successResponse = {
+        statusCode: 200,
+        devicename: request.body.newName,
+        groupname: newGroup.groupname,
+        messages: ["name and gruop set"],
```
mjs: group

api/V1/Device.js
```diff
@@ -239,4 +239,67 @@ module.exports = function(app, baseUrl) {
...
+      };
+
+      if (conflictingDevice != null) {
+        if (conflictingDevice.id == request.device.id) {
+          return responseUtil.send(response, successResponse);
+        } else {
+          return responseUtil.send(response, {
+            statusCode: 400,
+            messages: ["already a deice in group '"+newGroup.groupname+"' with the name '"+newName+"'"]
```
mjs: device

api/V1/Device.js
```diff
@@ -239,4 +239,67 @@ module.exports = function(app, baseUrl) {
...
+          });
+        }
+      }
+
+      if (await request.device.rename(newName, newGroup.id)) {
```
mjs: Can you move the check for conflicts to `Device.rename` please? It could throw a ClientError error if the name is already taken. 

It would also be more robust if you could do the check for a conflict into a SERIALIZABLE transaction with the update. That way there's no chance of concurrent renames or other updates messing things up.

- https://sequelize.org/master/manual/transactions.html#managed-transaction--auto-callback-
- https://sequelize.org/master/manual/transactions.html#isolation-levels

PR #250, opened 2019-08-08
Added device history table

CameronRP:
Added a device history table so can keep track of name and group changes of a device. Could be expanded in the future for location...

### DIFF
--- a/migrations/20190808050641-add-device-rename-history.js
+++ b/migrations/20190808050641-add-device-rename-history.js
@@ -2,25 +2,23 @@
 
 const util = require("../models/util/util");
 
-const tableName = "DeviceHistories";
+const tableName = "DeviceHistory";
 
 module.exports = {
-  up: (queryInterface, Sequelize) => {
-    return queryInterface.createTable(tableName, {
+  up: async (queryInterface, Sequelize) => {
+    await  queryInterface.createTable(tableName, {
       oldName: { type: Sequelize.STRING, allowNull: false },
       newName: { type: Sequelize.STRING, allowNull: false },
       oldGroupID: { type: Sequelize.INTEGER, allowNull: false },
       newGroupID: { type: Sequelize.INTEGER, allowNull: false },
       createdAt: { type: Sequelize.DATE, allowNull: false },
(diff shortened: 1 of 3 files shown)


migrations/20190808050641-add-device-rename-history.js
```diff
@@ -0,0 +1,26 @@
...
+      newName: { type: Sequelize.STRING, allowNull: false },
+      oldGroupID: { type: Sequelize.INTEGER, allowNull: false },
+      newGroupID: { type: Sequelize.INTEGER, allowNull: false },
+      createdAt: { type: Sequelize.DATE, allowNull: false },
+      updatedAt: { type: Sequelize.DATE, allowNull: false },
+    }).then(() => {
+      return util.addSerial(queryInterface, tableName);
+    }).then(() => {
+      return util.migrationAddBelongsTo(queryInterface, tableName, "Devices");
+    });
```
mjs: Why'd you go with `.then` instead of async and await?

migrations/20190808050641-add-device-rename-history.js
```diff
@@ -0,0 +1,26 @@
...
+  },
+
+  down: (queryInterface) => {
+    return queryInterface.sequelize.query('DROP TABLE "'+tableName+'" CASCADE');
```
mjs: I don't think the CASCADE here adds any value because no table refers to `DeviceHistories` (only the other way around).

migrations/20190808050641-add-device-rename-history.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const util = require("../models/util/util");
+
+const tableName = "DeviceHistories";
```
mjs: I wonder if `DeviceHistory` would be a better table name in this case

migrations/20190808050641-add-device-rename-history.js
CameronRP: Not too sure.. Will change that to use async and await.

mjs: I'm guessing you used an existing migration as starting point?

PR #257, opened 2019-08-22
added script to migrate old recordings tags to track tags

gferraro:


migrate-tags.js
```diff
@@ -0,0 +1,136 @@
+/*
+This is a one-off script to add type prefixes to object keys that don't have one.
```
mjs: Better make this correct :)

migrate-tags.js
```diff
@@ -0,0 +1,136 @@
...
+  "bird",
+  "kiwi",
+  "dog",
+  "leporidae",
+  "human",
+  "insect",
+  "pest",
+  "part"
+];
+Object.freeze(animals);
```
mjs: No need to change but you can also combine these like this:
```js
const animals = Object.freeze([...]);
```

migrate-tags.js
```diff
@@ -0,0 +1,136 @@
...
+    .parse(process.argv);
+
+  config.loadConfig(args.config);
+
+  const pgClient = await pgConnect();
+  const recordings = await getSingleTrackRecordings(pgClient);
+  let manualAnimals;
+  logger.info(`Found ${recordings.length} possible recordings`);
+  for (const [rId, trackId, oldTags] of recordings) {
+    manualAnimals = getManualAnimalTags(oldTags);
```
mjs: Why not just `const manualAnimal = ...`? No need to define this outside of the loop.

migrate-tags.js
```diff
@@ -0,0 +1,136 @@
...
+
+async function addTrackTag(client, rId, trackId, tag) {
+  logger.info(
+    `Adding Tag to ${trackId} for recording:${rId} animal:${tag.what}`
+  );
+  const text = `INSERT INTO "TrackTags" 
+  (what, confidence, data, automatic, "createdAt", "updatedAt", "TrackId", "UserId") 
+  VALUES($1, $2, $3, $4, $5, $6, $7, $8);`;
+  const values = [
+    `${tag.what}`,
```
mjs: There's no need for the string interpolation here is there? Just `tag.what,`

migrate-tags.js
```diff
@@ -0,0 +1,136 @@
...
+    tag.confidence,
+    '""',
+    tag.automatic,
+    tag.createdAt,
+    "`now()`",
+    trackId,
+    tag.taggerId
+  ];
+  await client.query(text, values).catch(e => logger.error(e.stack));
```
mjs: Does this means it just keeps going if there's an error? Is that really what you want?

migrate-tags.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+function getManualAnimalTags(oldTags) {
+  const manualAnimalTags = [];
+  let tag;
+  for (let i = 0; i < oldTags.length; i++) {
+    tag = oldTags[i];
+    if (
+      tag.automatic == false &&
```
mjs: `!tag.automatic`

migrate-tags.js
```diff
@@ -0,0 +1,136 @@
...
+      animals.includes(tag.what) &&
+      tag.condidence > MIN_CONFIDENCE
+    ) {
+      manualAnimalTags.push(tag);
+    }
+  }
+  return manualAnimalTags;
```
mjs: Ths function would be better written as something like:
```js
return oldTags.filter(tag => {
   return !tag.automatic && animals.includes(tag.what) && tag.confidence >= MIN_CONFIDENCE
});
```

Embrace the functional programming ideas :)

migrate-tags.js
```diff
@@ -0,0 +1,136 @@
...
+
+function getManualAnimalTags(oldTags) {
+  const manualAnimalTags = [];
+  let tag;
+  for (let i = 0; i < oldTags.length; i++) {
+    tag = oldTags[i];
+    if (
+      tag.automatic == false &&
+      animals.includes(tag.what) &&
+      tag.condidence > MIN_CONFIDENCE
```
mjs: confidence!

mjs: Better make it >=

migrate-tags.js
```diff
@@ -0,0 +1,136 @@
...
+      JOIN "Tracks" t on t."RecordingId" = r."id"
+      JOIN (
+        SELECT rSub."id", COUNT(t.*) as trackCount  
+        FROM "Recordings"  rSub 
+        JOIN "Tracks" t on t."RecordingId" = rSub."id" 
+        GROUP BY rSub."id" 
+        ) temp on temp.id = r.id
+      WHERE temp.trackCount = 1 AND
+      NOT EXISTS (
+                  SELECT 1 FROM "TrackTags" tag WHERE tag."TrackId" = t."id" AND tag."automatic"=false
```
mjs: You can also do `not tag.automatic` on the end here which reads better. No need to fix though.

migrate-tags.js
```diff
@@ -0,0 +1,136 @@
...
+      )  
+      AND r."additionalMetadata" IS NOT NULL
+      AND r."additionalMetadata" ? 'oldTags'
+      AND (r."additionalMetadata"->>'algorithm')::int >=41`,
```
mjs: I think if you use `->` instead of `->>` you'll get the value in it's actual type and not need to do the type case to int. `->>` means "get the value as text".

migrate-tags.js
```diff
@@ -76,23 +75,20 @@ async function addTrackTag(client, rId, trackId, tag) {
     trackId,
     tag.taggerId
   ];
-  await client.query(text, values).catch(e => logger.error(e.stack));
+  await client.query(text, values).catch(e => {
+    logger.error(e.stack);
+    process.exit(0);
```
mjs: this is fine but you could have also just let the exception bubble up with largely the same result (i.e. no `catch` at all)

PR #258, opened 2019-08-22
Allow CSV reports to be generated using a JWT argument

mjs:
- added /token endpoint which generates a short-lived JWT for a user
- these tokens are now accepted by the report endpoint
- lots of refactoring and cleanup of authentication code 

Still to come:
- further auth refactoring (it can be simplier and "simple is Good" when it comes to authentication
- limit temporary tokens to specific parts of the API

### DIFF
--- a/api/V1/AuthenticateUser.js
+++ b/api/V1/AuthenticateUser.js
@@ -17,10 +17,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 const jwt = require("jsonwebtoken");
-const config = require("../../config");
+const { body, oneOf } = require("express-validator/check");
+
+const auth = require("../auth");
 const responseUtil = require("./responseUtil");
 const middleware = require("../middleware");
-const { body, oneOf } = require("express-validator/check");
 
 module.exports = function(app) {
   /**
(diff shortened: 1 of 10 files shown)


api/V1/AuthenticateUser.js
```diff
@@ -57,21 +57,46 @@ module.exports = function(app) {
...
+   * generates a short-lived token which can be used as part of URLs.
+   *
+   * @apiUse V1UserAuthorizationHeader
+   * @apiSuccess {JSON} token JWT that may be used to call the report endpoint.
+   */
+  app.post(
+    "/token",
+    [auth.authenticateUser],
+    middleware.requestWrapper(async (request, response) => {
+      const token = auth.createEntityJWT(request.user, { expiresIn: 60 * 5 });
```
CameronRP: Do we really need this to be 5 minutes? The shorter the batter I would think.

CameronRP: Maybe add a "eventReport" type for signed JWTs so they are limited to just being used to get event reports. So you could have something like `auth.createEventReportJWT(user)` and that would return a JWT that can just be used for getting reports for the next minute?

mjs: I will make it shorter and I will add limiting in a follow up PR. There's a bunch of other auth related things I want to fix up.

CameronRP: OK. Sounds good.

PR #259, opened 2019-08-26
Added FileBytes and Content-Length to signedUrl

gferraro:


api/V1/File.js
```diff
@@ -132,10 +133,15 @@ module.exports = (app, baseUrl) => {
...
 
+      const s3Data = await util.getS3Object(file.fileKey).catch(err => {
+        return responseUtil.serverError(response, err);
+      });
+
       return responseUtil.send(response, {
         statusCode: 200,
         messages: [],
         file: file,
+        fileBytes: s3Data.ContentLength,
```
mjs: I wonder if `fileSize` would be a better name? `fileBytes` makes me think of the contents of the file.

api/V1/Recording.js
```diff
@@ -224,11 +226,20 @@ module.exports = (app, baseUrl) => {
...
+        cookedBytes,
+        cookedJWT
+      } = await recordingUtil.get(request);
+
       responseUtil.send(response, {
         statusCode: 200,
         messages: [],
         recording: recording,
+        rawBytes: rawBytes,
+        fileBytes: cookedBytes,
```
mjs: Same here: `rawSize` and `fileSize`

api/V1/util.js
```diff
@@ -119,4 +119,14 @@ function multipartUpload(keyPrefix, buildRecord) {
   };
 }
 
+async function getS3Object(fileKey) {
```
mjs: this function doesn't use `await` so the `async` here doesn't do anything

migrations/20190826020311-remove-file-size.js
```diff
@@ -0,0 +1,17 @@
+"use strict";
+
+module.exports = {
+  up: queryInterface => {
+    return Promise.all([
+      queryInterface.removeColumn("Recordings", "fileSize"),
```
mjs: There's also a `rawFileSize` column on Recordings

PR #261, opened 2019-08-28
Change rename to reregister

CameronRP:
Remove rename and replace with reregister.


migrations/20190828033536-add-saltid.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return Promise.all([
+      queryInterface.addColumn("Devices", "saltId", Sequelize.INTEGER),
```
mjs: Should the migration also pre-populate the saltId values?

test/test_device_reregister.py
```diff
@@ -0,0 +1,52 @@
...
+        new_name = "new_name_" + helper.random_id()
+        new_group = helper.make_unique_group_name(self, "new_group")
+        new_password = helper.random_password()
+        user1.create_group(new_group)
+
+        # should error if not changing name or group
+        with pytest.raises(Exception):
+            device1.reregister(new_name, new_group)
+
+        # fail movign to a group that doesn't exist
```
mjs: moving

models/Device.js
```diff
@@ -349,39 +357,33 @@ module.exports = function(sequelize, DataTypes) {
...
-            newGroupID: newGroup.id,
-            oldGroupID: this.getDataValue("GroupId"),
-            DeviceId: this.getDataValue("id")
-          },
-          { transaction: t }
-        );
+        await Device.update({
+          active: false
+        }, {
+          where: {saltId: this.getDataValue("saltId")},
```
mjs: I'm pretty sure you can just go `this.saltId`

models/Device.js
```diff
@@ -349,39 +357,33 @@ module.exports = function(sequelize, DataTypes) {
...
-            devicename: newName,
-            GroupId: newGroup.id
-          },
-          { transaction: t }
-        );
+        newDevice = await models.Device.create({
+          devicename: newName,
+          GroupId: newGroup.id,
+          password: newPassword,
+          saltId: this.getDataValue("saltId"),
```
mjs: same here

test/helper.py
```diff
@@ -170,3 +170,6 @@ def _check_admin_and_group_exist(self):
 
     def random_id(self, length=6):
         return "".join(random.choice(string.ascii_uppercase + string.digits) for _ in range(length))
+
+    def random_password(self, length=10):
+        return "".join(random.choice(string.ascii_uppercase + string.digits) for _ in range(length))
```
mjs: How about you make this a call to random_id with a higher length?

test/test_device_reregister.py
```diff
@@ -0,0 +1,52 @@
...
+        device2 = helper.given_new_device(self)
+
+        # can't have same name and group as another device
+        with pytest.raises(Exception):
+            device2.reregister(device1_new_name, device1_new_group, "password")
+
+        # can't reregister with user token
+        device2._deviceapi._auth_header = user1._userapi._auth_header
+        with pytest.raises(Exception):
+            device2.reregister(helper.random_id(), device1_new_group, "password")
```
mjs: This test is actually testing a variety of different cases. It would be clearer if that were each in their own test.

Don't worry about changing this now (unless you want to) - just something to keep in mind for the future.

PR #262, opened 2019-08-29
added confidence of tags to query

gferraro:


test/test_recordings_filter.py
```diff
@@ -59,3 +59,27 @@ def test_latitude_longitude_filter(self, helper):
         print("  Recordings from a normal query shoudl also be filtered", end="")
         rec = bob.query_recordings(filterOptions=json.dumps({"latLongPrec": 10}))[0]
         assert rec["location"]["coordinates"] == [20.00025, 20.00025]
+
+    def test_recordfing_query_fields(self, helper):
```
mjs: how about `test_recording_query_tag_fields`?

test/test_recordings_filter.py
```diff
@@ -59,3 +59,27 @@ def test_latitude_longitude_filter(self, helper):
...
+        larry_device = helper.given_new_device(self, "larrys_device", larry_group)
+        recording = larry_device.upload_recording()
+        track = larry.can_add_track_to_recording(recording)
+        larry.can_tag_track(track, automatic=False, what="possum")
+        recording.is_tagged_as(what="foo", detail="bar").by(larry)
+
+        response = larry.query_recordings(return_json=True)
+        for r in response["rows"]:
+            tags = r.get("Tags")
+            assert tags is not None
```
mjs: Instead of using get() and then checking for None you could just `tags = r["Tags"]` which will fail with a explantory `KeyError` if it fails. Same for the similar cases below.

test/test_recordings_filter.py
```diff
@@ -59,3 +59,27 @@ def test_latitude_longitude_filter(self, helper):
...
+            for tag in tags:
+                assert tag.get("confidence") != 0
+            tracks = r.get("Tracks")
+            assert tracks is not None
+            for track in tracks:
+                track_tags = track.get("TrackTags")
+                assert track_tags is not None
+                for track_tag in track_tags:
+                    assert track_tag.get("confidence") != 0
```
mjs: what about checking the other tag fields?

PR #265, opened 2019-09-03
Fix Recordings Query Count

gferraro:


api/V1/recordingUtil.js
```diff
@@ -63,6 +63,7 @@ async function query(request, type) {
     request.query.limit,
     request.query.order
   );
+  builder.query.distinct = true;
```
mjs: Does this have any effect on the query results themselves, or just the count?

gferraro: Looking at the query that is run it seems to just be the count, saw a few posts suggesting to do this online.

mjs: Cool, sounds good

PR #266, opened 2019-09-09
added device query endpoint and token access and ttl

gferraro:


api/V1/AuthenticateUser.js
```diff
@@ -22,6 +22,8 @@ const auth = require("../auth");
 const responseUtil = require("./responseUtil");
 const middleware = require("../middleware");
 
+const ttyTypes = Object.freeze({ short: 60, medium: 5 * 60, long: 30 * 60 });
```
mjs: I think you mean ttl (time to live) not tty (which typically means teletype or terminal type)

api/V1/AuthenticateUser.js
```diff
@@ -83,14 +85,26 @@ module.exports = function(app) {
...
+   * @apiParam {String} ttl short,mediam,long defining token expiry time
+   * @apiParam {JSON} access dictionary of access to different entities
+   *  e.g. {"devices":"r"}
    * @apiUse V1UserAuthorizationHeader
    * @apiSuccess {JSON} token JWT that may be used to call the report endpoint.
    */
   app.post(
     "/token",
-    [auth.authenticateUser],
+    [body("ttl").optional(), body("access").optional(), auth.authenticateUser],
```
mjs: FYI, I'm wondering about making `access` compulsory in a future PR (probably if/when I do that auth refactoring I mentioned yesterday). It will be safer to only ever give out tokens that are limited to a specific area of functionality.

Not for this PR though...

api/V1/Device.js
```diff
@@ -279,4 +279,51 @@ module.exports = function(app, baseUrl) {
       });
     })
   );
+
+  /**
+   * @api {post} /api/v1/devices/query Querys devices by group:name.
```
mjs: "Query devices by group and name"

api/V1/Device.js
```diff
@@ -279,4 +279,51 @@ module.exports = function(app, baseUrl) {
...
+   * @apiName query
+   * @apiGroup Device
+   * @apiDescription This call is to query all devices by groups or devicenames
+   *
+   * @apiUse V1DeviceAuthorizationHeader
+   *
+   * @apiParam {JSON} devices array of device obejcts [{devicename:"<name>",groupname:"<group>"}].
```
mjs: objects

mjs: You can use `@apiParamExample` for the example here (so that it's formatted nicely)

api/V1/Device.js
```diff
@@ -279,4 +279,51 @@ module.exports = function(app, baseUrl) {
...
+   * @apiParam {String} groups array of group names.
+   * @apiParam {String} operator sequelize operator to join devices and groups.
+   * by defualt $or "device in devices $OR device in groups"
```
mjs: default

api/V1/Device.js
```diff
@@ -279,4 +279,51 @@ module.exports = function(app, baseUrl) {
...
+
+   *
+   * @apiUse V1ResponseSuccess
+   * @apiUse V1ResponseError
+   */
+  app.post(
```
mjs: Why is it a POST when it's only querying? Shouldn't it be a GET?

api/V1/Device.js
```diff
@@ -279,4 +279,51 @@ module.exports = function(app, baseUrl) {
...
+    apiUrl + "/query",
+    [
+      body("username").optional(),
+      auth.authenticateAccess("user", { devices: "r" })
+    ],
+    middleware.requestWrapper(async function(request, response) {
+      if (
+        request.body.username &&
+        request.body.username != request.user.username
+      ) {
```
mjs: can you explain what this check is for?

mjs: also, `username` isn't documented in the API doc

api/V1/Device.js
mjs: please, no more sequelize leaking out of the API! I'm fine with a parameter for specifying this behaviour but we shouldn't be feeding input from the API straight into the ORM layer.

I know we do it elsewhere but I want to gradually get rid of those cases. Let's not add more.

api/auth.js
```diff
@@ -21,12 +21,14 @@ const jwt = require("jsonwebtoken");
...
 const models = require("../models");
-
 /*
  * Create a new JWT for a user or device.
  */
-function createEntityJWT(entity, options) {
+function createEntityJWT(entity, options, access) {
   const payload = entity.getJwtDataValues();
+  if (access) {
+    payload.access = access;
```
mjs: `access` is a decent name. I was struggling to come up with something for this when I was thinking about it earlier.

test/testuser.py
```diff
@@ -262,6 +261,11 @@ def get_active_devices(self):
                 active_devices.append(device)
         return active_devices
 
+    def query_devices(self, username=None, devices: List[TestDevice] = None, groups: List[str] = None):
```
mjs: Nice use of type annotations :) 
We don't have anything checking them at present (we would need to start using mypy) so they're purely for documentation (which is fine and worthwhile)

test/userapi.py
```diff
@@ -22,9 +24,21 @@ def register_as_new(self):
...
-        return self._check_response(response)
+    def token(self, access=None, set_token=False):
+        post_data = {}
+        if access is not None:
+            post_data["access"] = access
+
+        headers = self._auth_header.copy()
+        headers["Content-Type"] = "application/json"
+        response = requests.post(
+            urljoin(self._baseurl, "/token"), headers=headers, data=json.dumps(post_data)
```
mjs: If you do `json=post_data` instead of converting it yourself and passing it via `data` then requests will also set  Content-Type to application/json for you.

test/testuser.py
gferraro: Should keep cam happy if he ever sees it

PR #267, opened 2019-09-15
Restore the legacy /api/v1/recordings/:devicename POST endpoint 

mjs:
This is still be used by Sidekick in the wild so we need to support it for a little longer.

Also fixed sporadically failing device query tests.

### DIFF
--- a/api/V1/Recording.js
+++ b/api/V1/Recording.js
@@ -143,6 +143,34 @@ module.exports = (app, baseUrl) => {
   ]);
 
   /**
+   * @api {post} /api/v1/recordings/:devicename Legacy upload on behalf of
+   * @apiName PostRecordingOnBehalfLegacy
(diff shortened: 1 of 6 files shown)


api/V1/Recording.js
```diff
@@ -142,6 +142,35 @@ module.exports = (app, baseUrl) => {
...
+   *
+   * @apiUse V1ResponseSuccess
+   * @apiSuccess {Number} recordingId ID of the recording.
+   * @apiuse V1ResponseError
+   */
+  app.post(
+    apiUrl + "/:devicename",
+    [
+      auth.authenticateUser,
+      middleware.setGroupName(param),
```
CameronRP: Is this needed?

mjs: good point

PR #270, opened 2019-10-31
csv timezone is now set to New Zealand

gferraro:
also max query limit for recordings to 100,000
also addwed location to export


api/V1/recordingUtil.js
```diff
@@ -184,16 +184,22 @@ async function report(request) {
       r.type,
       r.Group.groupname,
       r.Device.devicename,
-      moment(r.recordingDateTime).format("L"),
-      moment(r.recordingDateTime).format("LTS"),
+      moment(r.recordingDateTime)
+        .tz("Pacific/Auckland")
```
mjs: move this to a const (and check if there's anywhere else that refers to it)

api/V1/recordingUtil.js
```diff
@@ -132,7 +132,9 @@ async function report(request) {
       "Device",
       "Date",
       "Time",
+      "Location",
```
mjs: This probably needs to be split out into lat and lon

config.js
```diff
@@ -6,6 +6,8 @@ const server = {
   loggerLevel: "info"
 };
 
+const timeZone = "Pacific/Auckland";
```
mjs: great idea to put it into config

api/V1/recordingUtil.js
```diff
@@ -187,12 +188,13 @@ async function report(request) {
       r.Group.groupname,
       r.Device.devicename,
       moment(r.recordingDateTime)
-        .tz("Pacific/Auckland")
+        .tz(config.timeZone)
         .format("YYYY-MM-DD"),
       moment(r.recordingDateTime)
-        .tz("Pacific/Auckland")
+        .tz(config.timeZone)
         .format("h:mm a"),
```
mjs: What about the seconds? 

You could just go with 24hr time here, e.g. 14:06:11

PR #272, opened 2019-11-06
added replace tag endpoint

gferraro:


api/V1/Recording.js
```diff
@@ -480,6 +480,71 @@ module.exports = (app, baseUrl) => {
     })
   );
 
+  /**
+   * @api {post} /api/v1/recordings/:id/tracks/:trackId/replaceTag/
+   * Adds or Replaced track tag based off:
```
mjs: Adds or replaces track tags:

models/Track.js
```diff
@@ -46,6 +46,30 @@ module.exports = function(sequelize, DataTypes) {
...
+      const existingAnimalTags = trackTags.filter(function(uTag) {
+        return !uTag.isAdditionalTag();
+      });
+
+      for (let i = 0; i < existingAnimalTags.length; i++) {
+        await existingAnimalTags[i].destroy();
+      }
+    }
+
+    await tag.save();
```
mjs: Can this be done inside a transaction to avoid problems with multiple replaceTag requests for a track coming in at around the same time?

PR #273, opened 2019-11-07
ability to query only on device

gferraro:


api/V1/Device.js
```diff
@@ -300,6 +299,8 @@ module.exports = function(app, baseUrl) {
    * @apiParam {String} groups array of group names.
    * @apiParam {String} operator to use. Default is "or".
    * Accepted values are "and" or "or".
+   * @apiSuccess {JSON} devices Array of devices which match the fully
```
mjs: sentence didn't get finished

models/Device.js
```diff
@@ -279,6 +279,7 @@ module.exports = function(sequelize, DataTypes) {
    */
   Device.queryDevices = async function(authUser, devices, groups, operator) {
```
mjs: It would be clearly if the various parts of this function were extracted into smaller methods. It's pretty long and hairy.

For this PR or later, I don't mind.

PR #274, opened 2019-11-11
added index to tracks and trackstags

gferraro:
changed recording query to query Tags and/or TracksTags tables when relevant.
also changed tests to add trackstags for animals instead of recording tags

### DIFF
--- a/migrations/20191111002157-track-index.js
+++ b/migrations/20191111002157-track-index.js
@@ -1,0 +1,12 @@
+"use strict";
+
+module.exports = {
+  up: async function(queryInterface) {
+    await queryInterface.addIndex("Tracks", ["RecordingId"]);
+    await queryInterface.addIndex("TrackTags", ["TrackId"]);
+  },
+  down: async function(queryInterface) {
+    await queryInterface.removeIndex("Tracks", ["RecordingId"]);
+    await queryInterface.removeIndex("TrackTags", ["TrackId"]);
+  }
+};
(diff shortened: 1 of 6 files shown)


migrations/20191111002157-track-index.js
```diff
@@ -0,0 +1,12 @@
...
+module.exports = {
+  up: async function(queryInterface) {
+    await queryInterface.addIndex("Tracks", ["RecordingId"]);
+    await queryInterface.addIndex("TrackTags", ["TrackId"]);
+  },
+  down: async function(queryInterface) {
+    await queryInterface.removeIndex("Tracks", ["RecordingId"]);
+    await queryInterface.removeIndex("TrackTags", ["TrackId"]);
+  }
+};
```
mjs: so these make a decent difference to the query plan?

models/Recording.js
```diff
@@ -580,6 +580,21 @@ module.exports = function(sequelize, DataTypes) {
 
     const humanSQL = 'NOT "Tags".automatic';
     const AISQL = '"Tags".automatic';
+    if (models.Tag.acceptableTags.includes(tagMode)) {
```
mjs: Is this right? It doesn't look like `tagMode` would ever be in `acceptableTags`.

migrations/20191111002157-track-index.js
gferraro: yep it uses these indexes which remove a cost  a cost of 20% due to searching via a hash prior

models/Recording.js
gferraro: when you serach on browse for cool that is put into the tagMode. I think we should change this so we sepearte recording Tags and TrackTags in the query

mjs: Separating the track and recording tags in the query makes a lot of sense.

PR #278, opened 2019-11-22
updated recording queries to return consistent results

gferraro:


models/Recording.js
```diff
@@ -534,44 +517,77 @@ module.exports = function(sequelize, DataTypes) {
...
+          [
+            Sequelize.fn(
+              "json_build_object",
+              "start_s",
+              Sequelize.literal(`"Tracks"."data"#>'{start_s}'`),
+              "end_s",
+              Sequelize.literal(`"Tracks"."data"#>'{end_s}'`)
+            ),
+            "data"
+          ]
```
mjs: What does it mean when you put an array in the `attributes` array like this?

test/test_recordings_filter.py
```diff
@@ -85,8 +85,14 @@ def test_recordfing_query_tag_fields(self, helper):
             assert tracks is not None
             for track in tracks:
                 track_tags = track.get("TrackTags")
```
mjs: This line belongs underneath the start_s and end_s asserts

models/Recording.js
gferraro: ['test_data', 'test'] e.g. which would translate too "select test_data as test"

test/test_recordings_filter.py
gferraro: > Have you checked that this works with all parts of the cacophony-browse that use these APIs?

Have run through with browse, shouldn't cause any problems as it's only returning more than it was before

PR #282, opened 2019-12-03
Added end user agreement

CameronRP:


api/V1/User.js
```diff
@@ -158,4 +164,25 @@ module.exports = function(app, baseUrl) {
       });
     })
   );
+
+  /**
+   * @api {get} api/v1/endUserAgreementVersion Get the latest end user agreement version
```
mjs: missing / before `api`

api/V1/User.js
```diff
@@ -158,4 +164,25 @@ module.exports = function(app, baseUrl) {
...
+   * @apiName EndUserAgreementVersion
+   * @apiGroup User
+   *
+   * @apiSuccess {Integer} euaVersion Version of the latest end user agreement.
+   * @apiUse V1ResponseSuccess
+   *
+   * @apiUse V1ResponseError
+   */
+  app.get(
+    baseUrl + "/endUserAgreementVersion",
```
mjs: I can imagine that we might end up having the EUA text stored in the API server so perhaps leave room for this by making the URL `/api/v1/endUserAgreement/latest`

Issue #207, opened 2019-05-12
Remove use of serialised JSON strings in request bodys if favour of actual JSON


jackodsteel:
Currently multiple endpoints expect a field in the request body to be a string serialised JSON object. This makes sense in `GET` requests where the query params must be strings, but in the bodys of `POST`/`PATCH` endpoints, we should either enforce or support JSON objects as opposed to their string representations.

In some cases, the request 

- [ ] `PATCH` `/api/v1/users` - data field in body
- [ ] `POST` `/api/v1/tags` - tag field in body
- [ ] `PATCH` `/api/v1/recordings/:id` - updates field in body
- [ ] `POST` `/api/v1/recordings/:id/tracks` - data and algorithm fields in body
- [ ] `POST` `/api/v1/recordings/:id/tracks/:trackId/tags` - data field in body
- [ ] `POST` `/api/v1/recordings/reprocess/multiple` - recordings field in body
- [ ] `POST` `/api/v1/schedules` - devices and schedule fields in body
- [ ] `/api/fileProcessing` - Multiple locations

In some cases these APIs only expect this single string, in which case the expected content in that JSON string should be come the new root values in the JSON body.
In other cases, there is other expected input in the body as well. In these cases the same field should be used, but its content should be the actual JSON object, not a string serialisation of it.

If single root field in body:
```
Current:
{
	"data": "{\"password\": \"a_new_password\", \"username\": \"a_new_username\"}"
}
New proposed:
{
	"password": "a_new_password",
	"username": "a_new_username"
}
```
If multiple root fields in body:
```
Current:
{
	"data": "{\"password\": \"a_new_password\", \"username\": \"a_new_username\"}",
	"otherFields": "blah"
}
New proposed:
{
	"data": {
		"password": "a_new_password",
		"username": "a_new_username"
	},
	"otherFields": "blah"
}
```

As an intermediate period we can still support both while downstream services switch. This would be done by checking the type of the field, if it's a string use the legacy implementation. If not then use the new. Once all services are migrated, remove legacy.

There is something to be noted in the priority of fixes. I've already made a PR to fix the `PATCH` `/api/v1/users` endpoint, as that's something that we could expect to be used more frequently than others. We should prioritise endpoints that will see more frequent use, especially by other developers if they so choose. Internal endpoints such as those related to file processing should still be fixed, but aren't as necessary.

mjs:
Thanks for writing this up. These rough edges would be nice to clean up.

Issue #199, opened 2019-05-02
Remove data field from PATCH users and instead use standard JSON input


jackodsteel:
Currently the PATCH /api/v1/users endpoint expects a JSON object containing a single "data" field, which contains a string which is supposed to decode to the expected JSON blob.

There's no reason to have this indirection, instead we should take the fields directly in JSON.

Current:
```
{
	"data": "{\"password\": \"a_new_password\", \"username\": \"a_new_username\"}"
}
```

New proposed:
```
{
	"password": "a_new_password",
	"username": "a_new_username"
}
```

As an intermediate period we can still support both while downstream services switch. This would be done by checking if there was a "data" field, if so use the legacy implementation. If not then use the new. Once all services are migrated, remove legacy.

mjs:
This is a nice improvement. :+1: 

cacophony-browse is the only consumer of this API that I'm aware of, so the transition period can be quite small.

PR #283, opened 2019-12-03
Add additional CORS Access-Control-Allow-Headers that are standard

jackodsteel:
This will allow Cacophony Browse to use JSON objects instead of url encoding.

The added headers are standard ones that are basically always enabled as seen below and on numerous other guides about CORS
https://enable-cors.org/server_expressjs.html

The one we really care about is Content-Type so we can send JSON.

#199 and #207 rely on this fix to get Cacophony Browse in spec as it currently used form encoding which is messier and less flexible.

### DIFF
--- a/Server.js
+++ b/Server.js
@@ -47,7 +47,7 @@ app.all("*", function(req, res, next) {
   );
   res.header(
     "Access-Control-Allow-Headers",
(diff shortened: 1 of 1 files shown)


mjs:
Great. thanks!

PR #285, opened 2019-12-04
Update parseJSON middleware to work for objects or strings

jackodsteel:
Also refactored the mapping of error messages to be simpler/functional style.

This addresses the serverside fixes for all of #207

### DIFF
--- a/api/customErrors.js
+++ b/api/customErrors.js
@@ -68,16 +68,11 @@ class ValidationError extends CustomError {
     super();
     this.statusCode = 422;
     this.errors = errors;
-
-    const mappedErrors = this.errors.mapped();
-    const errorStrings = [];
-    for (const name in mappedErrors) {
-      const error = mappedErrors[name];
-      if (typeof error.msg == "string") {
-        errorStrings.push(error.msg);
-      }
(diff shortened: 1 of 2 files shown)


api/customErrors.js
```diff
@@ -68,16 +68,11 @@ class ValidationError extends CustomError {
...
-      if (typeof error.msg == "string") {
-        errorStrings.push(error.msg);
-      }
-    }
-    this.message = errorStrings.join("; ");
+    this.message = this.errors
+      .array()
+      .filter(error => typeof error.msg === "string")
+      .map(error => error.msg)
+      .join("; ");
```
mjs: much better

jackodsteel:
> Are there parts of the docs for the API handling methods that need to be updated to reflect this improvement?

Not that I can see, because the existing docs are misleading in the fact that they say something like
```
* @apiparam {JSON} tag Tag data in JSON format.
```
Which implies (at least to me) that it's just JSON, but actually before this fix it was a JSON _string_ that was required.
That's how I found the issue initially actually. We're just bringing the implementation in line with what the docs say.

Issue #180, opened 2019-03-21
Deprecate generalised "where" query parameter


mjs:
Instead of accepting sequelize query expressions the recordings search API should take the different search critiera as query parameters.

This is more robust and gives us flexibility in terms of moving away from sequelizejs.

It also makes things like this easier: https://github.com/TheCacophonyProject/cacophony-browse/issues/150

PR #293, opened 2019-12-18
Fix 38 & 130, delete recording files from S3 when recording is deleted, and dont return JWTs if files don't exist

### LINKED ISSUE #26: Include failed file processing jobs through the API request.
When a file processing job is requests, if the file-processing fails and doesn't respond the file processing job should be included when doing an API request so it can be processed.
This can be detected by if the file processing start time is over an hour old, as a file should not take more than an hour to be processed.

### LINKED ISSUE #38: Delete file both raw and processed when deleting a recording.
Currently the recording just gets deleted from the database but the files should also be deleted from the object store too. To avoid keeping the client waiting, it might be nice to return back to HTTP client once the database deletion has been completed and then continue with the object store removals afterwards.

### LINKED ISSUE #130: JWT should not returned if recording not available
Especially for the processed recordings, if recording content isn't available, a JWT shouldn't be returned for it. Current a JWT with an empty key is returned.

jackodsteel:
Fixes #38, Fixes #130

#38 
Update Recording delete to also delete the S3 files in the background.
Will log a warning if it wasn't deleted but will still return 200 before that results and no matter the status of the S3 deletions.
Didn't add a test for this but you can verify with minio/mc, create recording, GET it and note the rawFileKey, configure mc if you haven't already: `mc config host add s3 http://localhost:9001 minio miniostorage `, then query the file such as `mc ls s3/cacophony/raw/2019/12/19/c4b2b587-a463-45e6-a499-40b95bd25c1b`, should return the file and size. Delete the file through UI. Run the mc ls command again and the it should say file not found.

#130 
Update recording JWTs to only apply if the file exists
Extracted out a method to get the filesize of files in S3 to simplify code.
Consumers will have to check for the existence of the field now, although Cacophony Browse currently seems to behave the same with or without this change.
Added a test for this.

### DIFF
--- a/api/V1/recordingUtil.js
+++ b/api/V1/recordingUtil.js
@@ -292,6 +292,9 @@ async function get(request, type) {
     data.rawSize = await util.getS3ObjectFileSize(recording.rawFileKey);
   }
 
+  delete data.recording.rawFileKey;
+  delete data.recording.fileKey;
+
   return data;
 }
 
--- a/api/V1/recordingUtil.js
+++ b/api/V1/recordingUtil.js
@@ -23,6 +23,7 @@ const urljoin = require("url-join");
 
 const { ClientError } = require("../customErrors");
 const config = require("../../config");
+const log = require("../../logging");
 const models = require("../../models");
 const responseUtil = require("./responseUtil");
 const util = require("./util");
@@ -283,17 +284,24 @@ async function delete_(request, response) {
     request.user,
     request.params.id
   );
-  if (deleted) {
-    responseUtil.send(response, {
-      statusCode: 200,
-      messages: ["Deleted recording."]
-    });
-  } else {
-    responseUtil.send(response, {
+  if (deleted === null) {
+    return responseUtil.send(response, {
       statusCode: 400,
       messages: ["Failed to delete recording."]
     });
   }
+  if (deleted.rawFileKey) {
+    util.deleteS3Object(deleted.rawFileKey)
+      .catch(err => { log.warn(err); });
+  }
+  if (deleted.fileKey) {
+    util.deleteS3Object(deleted.fileKey)
+      .catch(err => { log.warn(err); });
+  }
+  responseUtil.send(response, {
+    statusCode: 200,
+    messages: ["Deleted recording."]
+  });
 }
 
 function guessRawMimeType(type, filename) {
--- a/api/V1/recordingUtil.js
+++ b/api/V1/recordingUtil.js
@@ -23,7 +23,6 @@ const urljoin = require("url-join");
 
 const { ClientError } = require("../customErrors");
 const config = require("../../config");
-const log = require("../../logging");
 const models = require("../../models");
 const responseUtil = require("./responseUtil");
 const util = require("./util");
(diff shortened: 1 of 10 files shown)


api/V1/recordingUtil.js
```diff
@@ -246,85 +246,62 @@ async function get(request, type) {
...
+      filename: recording.getFileName(),
+      mimeType: recording.fileMimeType
+    },
+    config.server.passportSecret,
+    { expiresIn: 60 * 10 }
+    );
+    data.cookedSize = await util.getS3ObjectFileSize(recording.fileKey);
   }
 
-  delete recording.rawFileKey;
```
jackodsteel: I've removed this line, as it's still being returned currently because the `handleLegacyTagFieldsForGetOnRecording` method recreates the object with that key back in, so this deletion is currently pointless.

Do we actually want/care about this being deleted, and if so are there other keys that should also be deleted?

mjs:
Thanks for this. Can you remove the changes for #26 from this PR? I'm not convinced that continually retrying a recording that has failed to process is useful. It's almost certainly going to keep failing as the reason for failure is usually bug in the processing backend or a problem with the recording's data. Both of these are things need human intervention to handle and I'd prefer we generated alerts instead (better observability and alerting is something we're currently working towards).

api/V1/recordingUtil.js
mjs: Good catch! 

We don't really want `rawFileKey` and `fileKey` to leak out of the API as they are internal details only. If it's easy, could you arrange things so they don't leak out here, or file an issue if it's going to be a big change?

jackodsteel: I've re-added the delete down the bottom where it'll now have the correct effect. Still feels a little prone to leakage as it relies on those delete statements at the end, but it's back to as it was at least.

test/test_file_processing.py
```diff
@@ -254,7 +254,7 @@ def test_reprocess_recording(self, helper, file_processing):
 
         # Now finalise processing.
         file_processing.put(recording, success=True, complete=True, new_object_key="some_key")
-        check_recording(admin, recording, processingState="FINISHED", fileKey="some_key")
+        check_recording(admin, recording, processingState="FINISHED")
```
mjs: Ahh... the file processing API is an internal API which needs to include fileKey and rawFileKey. if `get()` in recordingUtil.js is used for both places then the field removal needs to happen at a higher layer.

Sorry that I missed this detail.

mjs: Actually, never mind. I should have checked the code first. I see the file processing API gets the recording in a different way which still includes fileKey and rawFileKey.

Is there some other way that this test can confirm that the correct object key was set in the DB? The only approach I can think of is to actually upload some data to the object store at some key, set the recording to use that key and then download the recording to ensure the expected data is returned. 

Or, have a peek directly in the DB.

Perhaps not worth it...

test/test_recordings.py
```diff
@@ -18,3 +18,17 @@ def test_unprocessed_recording_doesnt_return_processed_jwt(self, helper):
...
+        bobsGroup = helper.make_unique_group_name(self, "bobs_group")
+        bob.create_group(bobsGroup)
+        bobsDevice = helper.given_new_device(self, "bobs_device", bobsGroup)
+        recording = bobsDevice.upload_recording()
+
+        print("And then fetches it")
+        recording_response = bob.get_recording(recording)
+
+        print("  The recording response should not contain the rawFileKey")
+        assert "rawFileKey" not in recording_response
```
mjs: and fileKey too?

test/test_file_processing.py
jackodsteel: See update, I've used the where param to include it as part of the query, which is the only way I can think of (although if you should be able to query based on fileKey is another issue for #180)

jackodsteel:
Should be good to go now @mjs :)

test/test_file_processing.py
```diff
@@ -254,7 +256,10 @@ def test_reprocess_recording(self, helper, file_processing):
 
         # Now finalise processing.
         file_processing.put(recording, success=True, complete=True, new_object_key="some_key")
-        check_recording(admin, recording, processingState="FINISHED")
+        query_result = admin.query_recordings(
+            where={"id": recording.id_, "fileKey": "some_key", "processingState": "FINISHED"}
+        )
+        assert len(query_result) == 1
```
mjs: This is a great way of testing that the fileKey is set.

PR #295, opened 2019-12-20
Typescriptify

hardiesoft:
I'm about ready for this to get put onto api-test.cacophony.org.nz alongside the 'tagging-ui' branch that integrate these changes on the front end.

One outstanding question I have for the api side of things is: how should we build the typescript files?  When developing locally I think it's reasonable to expect the developer to have some typescript watch job compiling the files as changes are made, but there probably needs to be some line in one of the build scripts that runs "tsc" - which should compile all the typescript sources as defined in tsconfig.json.

Thoughts?

Also, I realise there are a lot of changes going on here, and there is future work to be done around extracting common type definitions that the front-end and back-end both use into a third shared repo.  Let me know if there's anything I can do to help make life easier for whoever is reviewing.

The typings I've added do help me a lot to be able to navigate to definitions of things and jump around the code-base much more easily, but since Sequelize does so much dynamic monkey patching at runtime, it's never going to be as helpful as I'd like.


hardiesoft:
Just to be clear: CI will fail because the typescript files need to be compiled, and I haven't added a step for that yet :-)

mjs:
Thanks for this. I'm super excited the benefits of TypeScript.

The current developer workflow is to start the API server inside a docker container using the provided script. Unless `--isolate` is used the server will automatically restart (using nodemon). 

Now that TypeScript compilation needs happen could we have nodemon both run tsc and then restart the server when a code change is detected? Something like this: https://stackoverflow.com/a/40093215

Reviewing the code now..

mjs:
This approach looks to be superior (using `tsc --watch` not notice changes instead of nodemon): https://stackoverflow.com/a/47305304

passportConfig.ts
```diff
@@ -1,10 +1,11 @@
...
-const ExtractJwt = require("passport-jwt").ExtractJwt;
-const AnonymousStrategy = require("passport-anonymous");
-const models = require("./models");
-const config = require("../../config");
+import { ExtractJwt, Strategy as JwtStrategy } from "passport-jwt";
+import AnonymousStrategy from "passport-anonymous";
+import models from "./models";
 
-module.exports = function(passport) {
+// @ts-ignore - Assume that this references some out of repo config?
```
mjs: isn't this referring to config.js in the root of the repo?

hardiesoft:
Thanks for all this feedback, I should be able to address it all later in the week.

api/V1/index.ts
```diff
@@ -16,20 +16,25 @@ You should have received a copy of the GNU Affero General Public License
...
+export default function(app: Application) {
+  const excludedFiles = [
+    "index.js",
+    "util.js",
+    "responseUtil.js",
+    "recordingUtil.js",
+    "apidoc.js"
+  ];
+  // Filter out files that are not added to app directly, and filter out typescript versions of files.
+  // TODO(jon): This could be a great place to transpile any typescript files?
```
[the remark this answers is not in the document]
hardiesoft: Yes, agree that something around tsc --watch is the best solution, assuming we make all files .ts and not .js from now on.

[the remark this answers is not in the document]
mjs: I think we should aim to make all the source be TypeScript

passportConfig.ts
hardiesoft: Would that not be just `import config from "./config` rather than `import config from "../../config"`?

hardiesoft:
Hi, I'm not quite sure what I've done to make it so that it can't find the container running - I'm don't know docker well enough to work out what's going on in the Travis logs - probably something simple that I'm missing?  Any ideas?

passportConfig.ts
mjs: Right you are. I've just looked more closely and it seems like passportConfig isn't used at all. The function names in it aren't referenced anywhere and the tests pass when the file is removed. 

Please delete it in your PR.

mjs:
> Hi, I'm not quite sure what I've done to make it so that it can't find the container running - I'm don't know docker well enough to work out what's going on in the Travis logs - probably something simple that I'm missing? Any ideas?

The problem is that TypeScript compilation is required before the DB migrations can run, both for the config file and model definitions and supporting code.

I'm sorting this out now, as well as making troubleshooting this kind of thing easier in the future.

mjs:
There were a number of things broken which I've now fixed - see the commits I've added to this PR.

The only thing I've noticed that is potentially questionable is that `config/app_test_default.ts` is a TypeScript file. It only gets used as JavaScript so why not just keep it in the repo that way?

That can be changed later though so if the TravisCI checks pass I'm going to merge this.

mjs:
Stuff I'd like to see in future PRs:
- turning on strictNullChecks
- allowing trailing commas in multiline list expressions

PR #296, opened 2020-01-09
added migration to rename false positives

gferraro:
migrationg to rename false positive tracks tags to false-positive


mjs:
This needs to be updated to now that all the source is in TypeScript.

PR #299, opened 2020-01-15
Fixes bulk tagging api grabbing archived tracks.

hardiesoft:


config/app_test_default.js
```diff
@@ -1,7 +1,8 @@
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
```
mjs: Is this supposed to stay here?

hardiesoft: It was just part of the transpilation from the .ts version to the .js version - I will remove.

PR #300, opened 2020-01-21
Restore GroupUsers to unbreak groups admin view

hardiesoft:
I stopped group users from being bundled with getting groups, but it turns out some front-end views were relying on that.  Put it back for now and explore better solutions in the future?


mjs:
https://trello.com/c/3QD8F0FE/311-clean-up-group-and-user-apis

PR #306, opened 2020-02-05
Add system err event

CameronRP:
- Make the `devices` field always set when getting schedule. Fixes a error in cacophony-browse-old where you couldn't add a schedule to a device that didn't already have a schedule set.

- Specify sequelise to use equals when comparing the details JSONB for a event. 
When not set this is an example of the SQL generated 
`AND (CAST(("DetailSnapshot"."details"#>>\'{foo}\') AS DOUBLE PRECISION) = ARRAY[1,2,3] AND CAST(("DetailSnapshot"."details"#>>\'{test2}\') AS DOUBLE PRECISION) = 2)`
Resulting in `SequelizeDatabaseError: operator does not exist: double precision = integer[]`

### DIFF
--- a/api/V1/Schedule.ts
+++ b/api/V1/Schedule.ts
@@ -142,6 +142,10 @@ async function getSchedule(device: any, response: Response, user = null) {
     resData.devices = await models.Device.onlyUsersDevicesMatching(user, {
       ScheduleId: device.ScheduleId
     });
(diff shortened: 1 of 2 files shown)


CameronRP:
https://github.com/TheCacophonyProject/event-reporter/pull/18 can be merged after this is. Requires the new event type.

PR #310, opened 2020-03-03
Add user event upload

CameronRP:
Users can add events on behalf of devices same way that they can add recordings for devices.

### DIFF
--- a/api/V1/Event.ts
+++ b/api/V1/Event.ts
@@ -31,20 +31,16 @@ export default function(app: Application, baseUrl: string) {
    * @api {post} /api/v1/events Add new events
    * @apiName Add Event
    * @apiGroup Events
-   * @apiDescription This call is used to upload new events.   Events
-   * details are decided by user and can be specified by JSON or using an
-   * existing id.
+   * @apiDescription This call is used to upload new events.
+   * The event can be described by specifying an existing eventDetailId or by
+   * the 'description' parameter.
    *
    * `Either eventDetailId or description is required`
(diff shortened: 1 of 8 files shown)


api/V1/Event.ts
```diff
@@ -54,55 +55,45 @@ export default function(app: Application, baseUrl: string) {
...
+   * @api {post} /api/v1/events/device/:deviceID Add new events on behalf of device
+   * @apiName AddEventOnBehalf
+   * @apiGroup Events
+   * @apiDescription This call is used to upload new events on behalf of a device.
+   * Events details are decided by user and can be specified by JSON or using an
+   * existing id.
+   *
+   * `Either eventDetailId or description is required`
+   * - eventDetailsId: id
+   * - description:
```
gferraro: should this be under @apiParam ?
and could use the @apiParamExample tag

api/V1/Event.ts
```diff
@@ -54,55 +55,45 @@ export default function(app: Application, baseUrl: string) {
...
-        eventsAdded: count,
-        eventDetailId: detailsId
-      });
-    })
+  /**
+   * @api {post} /api/v1/events/device/:deviceID Add new events on behalf of device
+   * @apiName AddEventOnBehalf
+   * @apiGroup Events
+   * @apiDescription This call is used to upload new events on behalf of a device.
+   * Events details are decided by user and can be specified by JSON or using an
```
gferraro: Could this be worded a bit better?

PR #311, opened 2020-03-05
make devicename+group unique

gferraro:
remove unique devicename index, and make index on name + group
also change username, email, groupname and devicename to be case insensitive


CameronRP:
Looking good. Can you check that the migration would work on the test and prod API. I guess if we currently have devices "device1" and "Device1" the migration won't work?

PR #312, opened 2020-03-09
Species classification

hardiesoft:
Closes https://trello.com/c/vjGSxQzX

I've just added a CSV column to the export for species classification data, but there are also some additional type definitions for various types of recording "additionalMetadata".

Finally, I've tweaked the run.py script to work on Windows.


hardiesoft:
Actually, maybe going to hold off on this, since we should probably move the species classifier tags to be TrackTags, the way Tim has done things for his Morepork tagging.

PR #316, opened 2020-03-22
Cacophony index integration

hardiesoft:
This PR adds two new API endpoints to a device.
One to get an "average" cacophony index for the device from a given point in time stretching back over a given window, and another to give a 24hr histogram of averages for a device, again from a given point in time, and within a given window.

There are currently no API tests added for this functionality, since we'd also need to make sure our tests add recordings with cacophony-index data, and I'm not sure what's the best way to go about this currently.


api/V1/Device.ts
```diff
@@ -344,4 +344,104 @@ export default function(app: Application, baseUrl: string) {
...
+   * @apiGroup Device
+   * @apiDescription Get a single number Cacophony Index
+   * for a given device.  This number is the average of all the Cacophony Index values from a
+   * given time (defaulting to 'Now'), within a given timespan (defaulting to 3 months)
+   *
+   * @apiUse V1UserAuthorizationHeader
+   *
+   * @apiParam {Integer} deviceId ID of the device.
+   * @apiParam {String} from ISO8601 date string
+   * @apiParam {String} window-size length of rolling window in hours.  Default is 2160 (90 days)
```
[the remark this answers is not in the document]
gferraro: I dont think this is rolling.

models/Device.ts
```diff
@@ -351,6 +363,78 @@ export default function(
...
+
+  Device.getCacophonyIndexHistogram = async function(
+    authUser,
+    deviceId,
+    from,
+    windowSize
+  ) {
+    windowSize = Math.abs(windowSize);
+    // We need to take the time down to the previous hour, so remove 1 second
+    const date = Math.ceil(from.getTime() / 1000);
```
gferraro, line 409: again this is a timestamp rather than a date?

test/test_device_cacophony_index.py
```diff
@@ -0,0 +1,124 @@
...
+        cacophony_index = johnny.get_cacophony_index_histogram_for_device(self.get_device())
+        assert len(cacophony_index["cacophonyIndex"]) == 24
+        cacophony_index_admin = admin.get_cacophony_index_histogram_for_device(self.get_device())
+        assert len(cacophony_index_admin["cacophonyIndex"]) == 24
+
+        now = self.now()
+        cacophony_index_from_twelve_hours_ago = johnny.get_cacophony_index_histogram_for_device(
+            self.get_device(), js_iso_format_with_utc(now - timedelta(hours=12)), 12,
+        )
+        assert len(cacophony_index_from_twelve_hours_ago["cacophonyIndex"]) == 12
```
gferraro, line 109: could assert that the index values are all over 0 or something

.travis.yml
```diff
@@ -21,8 +21,11 @@ script:
  - black --diff --check ./test
  - ./wait-for-api || { docker ps; docker logs cacophony-api; exit 1; }
 # Apply migration undo and redo to test undo migration
+ - docker exec -it cacophony-api date
```
gferraro, line 24: is this left overs from testing?

gferraro:
> This PR adds two new API endpoints to a device.
> One to get an "average" cacophony index for the device from a given point in time stretching back over a given window, and another to give a 24hr histogram of averages for a device, again from a given point in time, and within a given window.
> 
> There are currently no API tests added for this functionality, since we'd also need to make sure our tests add recordings with cacophony-index data, and I'm not sure what's the best way to go about this currently.

api/V1/Device.ts
```diff
@@ -344,4 +344,104 @@ export default function(app: Application, baseUrl: string) {
...
+        .toInt()
+        .optional(),
+      auth.authenticateUser
+    ],
+    middleware.requestWrapper(async function(request, response) {
+      const cacophonyIndex = await models.Device.getCacophonyIndexHistogram(
+        request.user,
+        request.params.deviceId,
+        request.query.from || new Date(), // Get the current cacophony index
+        typeof request.query["window-size"] === "number"
```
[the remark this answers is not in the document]
hardiesoft: Because it's an optional param it can also be undefined, which is what we're checking for here.

api/V1/Device.ts
```diff
@@ -344,4 +344,104 @@ export default function(app: Application, baseUrl: string) {
...
+        .toInt()
+        .optional(),
+      auth.authenticateUser
+    ],
+    middleware.requestWrapper(async function(request, response) {
+      const cacophonyIndex = await models.Device.getCacophonyIndex(
+        request.user,
+        request.params.deviceId,
+        request.query.from || new Date(), // Get the current cacophony index
+        typeof request.query["window-size"] === "number"
```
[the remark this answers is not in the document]
hardiesoft: Because it's an optional param it can also be undefined, which is what we're checking for here.

test/test_device_cacophony_index.py
hardiesoft: > 
> 
> Looks good a few bits and pieces that should be tidied up.
> Also any reason for not using the Sequelize query format? we are using it everywhere, so good to use it to keep things consistent.

The main reason I am not using Sequelize query format is that I have not the slightest idea how I would start to express a query like this in sequelize format.  Do you know how you'd do it?

PR #317, opened 2020-04-07
Visits page

gferraro:
Moved visit generation into the api
Added endpoint for queryVisits
Added report option for visits


test/track.py
```diff
@@ -11,9 +11,11 @@ class Track:
     tags = attr.ib(factory=list)
 
     @classmethod
-    def create(cls, recording):
+    def create(cls, recording, start_s=10):
         "Make a TestTrack with some plausible data."
-        return cls(None, recording, data={"foo": [[1, 2], [3, 4]], "start_s": 10, "end_s": 20})
+        if start_s is None:
```
hardiesoft, line 16: Why have this and also have the parameter default be 10?  Genuine question, I'm a python noob, and there may be a good reason?

test/track.py
```diff
@@ -11,9 +11,11 @@ class Track:
...
+            start_s = 10
+        return cls(None, recording, data={"foo": [[1, 2], [3, 4]], "start_s": start_s, "end_s": start_s + 10})
```
hardiesoft, line 18: Is this actually plausible data for a track?  Don't we have something more formally defined?  Is the `foo` property just freeform data, or something else?

api/V1/recordingUtil.ts
```diff
@@ -525,6 +537,386 @@ async function updateMetadata(recording: any, metadata: any) {
...
+      .format("HH:mm:ss"),
+    "",
+    "",
+    "",
+    audioPlayed,
+    ""
+  ]);
+}
+
+// any visits whcih have started more than the visit interval from firstStart are marked as complted
```
hardiesoft, line 901: clean up your typos :-)

api/V1/recordingUtil.ts
```diff
@@ -525,6 +537,386 @@ async function updateMetadata(recording: any, metadata: any) {
...
+          audioTime = moment(audioEvent.dateTime);
+        }
+        audioBaitBefore = audioTime && audioTime.isAfter(event.start);
+      }
+      addEventRow(out, visit, event, recordingUrlBase);
+    }
+    if (audioEvent) {
+      audioEvents.push(audioEvent);
+    }
+    for (let i = audioEvents.length - 1; i >= 0; i--) {
```
hardiesoft, line 831: Could be clearer as `for (const audioEvent of audioEvents.reverse()) {` ?

api/V1/recordingUtil.ts
```diff
@@ -525,6 +537,386 @@ async function updateMetadata(recording: any, metadata: any) {
...
+  }
+
+  let queryOffset = 0;
+  // mark all as complete
+  if (gotAllRecordings) {
+    incompleteVisits.forEach(elem => {
+      elem.incomplete = false;
+    });
+
+    visits.splice(visits.length, 0, ...incompleteVisits);
```
hardiesoft, line 676: This looks like it should just be `visits.push(...incompleteVisits)`

test/track.py
gferraro: I'm guessing this is there, just to test that any data can be entered in there. there are probably other tests that rely on this.
Should really be removed and use more expected data like start_s and end_s

test/track.py
gferraro: No good reason, left overs from dev

hardiesoft:
@gferraro Testing with the latest changes, I get another 500 error when I try to export the report CSV:
`error:  TypeError: Cannot read property 'tz' of undefined
    at reportDeviceVisits (/app/api/V1/recordingUtil.js:544:36)
    at reportVisits (/app/api/V1/recordingUtil.js:575:17)
    at <anonymous>
`.

Maybe add some more type-checking to the functions in recordingUtil so that you know you're handling cases where some of the arguments are undefined?

api/V1/Visits.ts
```diff
@@ -112,18 +112,21 @@ class DeviceVisits {
...
-        visit.incomplete = !complete;
-        visit.queryOffset = queryOffset;
-        visits.push(visit);
+    for (const track of tracks) {
+      const event = this.calculateTrackTagEvent(rec, track);
+      if (event == null) {
+        continue;
+      }
+      this.updateSummary(event);
+      if (isVisit(event)) {
```
hardiesoft, line 125: Because you're using classes for Visit and VisitSummary, if you want to find out the type of one of these objects you can just do `if (event instanceof Visit) { ... }`, which is a bit more robust than checking for the existence of a particular property on the object.

gferraro:
> @gferraro Testing with the latest changes, I get another 500 error when I try to export the report CSV:
> `error: TypeError: Cannot read property 'tz' of undefined at reportDeviceVisits (/app/api/V1/recordingUtil.js:544:36) at reportVisits (/app/api/V1/recordingUtil.js:575:17) at <anonymous> `.
> 
> Maybe add some more type-checking to the functions in recordingUtil so that you know you're handling cases where some of the arguments are undefined?

Forget that booleans don't default to false in js, I guess thats why the scritnullcheck didn't pick it up.
I probably should create a test for the report, was just feeling lazy, also the fields, layout is sure to change.

api/V1/Recording.ts
```diff
@@ -177,6 +179,43 @@ export default (app: Application, baseUrl: string) => {
...
+   * @apiUse V1ResponseSuccessQuery
+   * @apiUse V1ResponseError
+   */
+  app.get(
+    apiUrl + "/visits",
+    [auth.authenticateUser, ...queryValidators],
+    middleware.requestWrapper(
+      async (request: e.Request, response: e.Response) => {
+        const result = await recordingUtil.queryVisits(
+          (request as unknown) as RecordingQuery
```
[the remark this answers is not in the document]
hardiesoft: This comment still stands.

api/V1/Visits.ts
```diff
@@ -0,0 +1,520 @@
...
+    } else {
+      this.eventCount += 1;
+    }
+
+    if (!this.startTime) {
+      this.startTime = visit.start;
+      this.endTime = visit.end;
+    } else {
+      if (this.startTime > visit.startTime) {
+        this.startTime = visit.STartTime;
```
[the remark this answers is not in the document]
hardiesoft: Typo still there - would be easier to spot if you added typing info to the visit argument.

api/V1/Visits.ts
```diff
@@ -0,0 +1,520 @@
...
+      rec,
+      track,
+      tag,
+      this.firstVisit.what != tag.what
+    );
+    this.addAudioFileIds(newEvent);
+    return newEvent;
+  }
+
+  unkownIsPartOfPreviousVisit(tag: TrackTag, end: Moment): boolean {
```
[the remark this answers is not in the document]
hardiesoft: Typo still exists

api/V1/Visits.ts
```diff
@@ -0,0 +1,537 @@
...
+class VisitEvent {
+  wasUnidentified: Boolean;
+  recID: number;
+  recStart: Moment;
+  trackID: number;
+  confidence: number;
+  start: Moment;
+  end: Moment;
+  audioBaitDay: boolean;
+  audioBaitEvents: any[];
```
hardiesoft, line 456: Should be Event[]

api/V1/Visits.ts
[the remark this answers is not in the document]
hardiesoft: I think this is the only outstanding issue on this PR.

PR #318, opened 2020-05-11
added script to send email about service errors in the last 24 hours

gferraro:


api/V1/systemError.ts
```diff
@@ -0,0 +1,365 @@
+// Fuzzily checks that error logs are similar
```
hardiesoft, line 1: It feels like you could replace most of this code with a [levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) implementation.  There are at least a couple of implementations on npm.

report-service-errors.ts
```diff
@@ -0,0 +1,107 @@
+import config from "./config";
+const winston = require("winston");
+import eventUtil from "./api/V1/eventUtil";
+// import models from "./models";
+import moment, { Moment } from "moment";
+import { ServiceErrorMap } from "./api/V1/systemError";
+
+var nodemailer = require("nodemailer");
```
hardiesoft, line 8: Maybe stick to import foo from bar syntax?

models/Event.ts
```diff
@@ -29,19 +29,28 @@ export interface Event extends Sequelize.Model, ModelCommon<Event> {
...
+export interface QueryOptions {
+  eventType: string;
+  admin: boolean;
+  useCreatedDate: boolean;
+}
+
 export interface EventStatic extends ModelStaticCommon<Event> {
   query: (
     user: User,
     startTime: string,
```
hardiesoft, line 44: startTime, endTime, deviceId, offset and limit are all marked as optional at the express layer, but here the type signature assumes they're always defined.  You might be missing some handling of cases where they are actually undefined?  Change to i.e. `number | undefined` and see what edge cases you need to handle.

api/V1/systemError.ts
gferraro: Not sure that any of them would do it like thist. Want to match per word and get substrings not with substitutions and not character based. I think this works best for logs, as we rarely or never should have words that are spelt differently in 2 logs

hardiesoft: Sorry, I thought the intention was just to get a fuzzy match metric for two strings.
If the use case is more nuanced than that I'd probably want to see some reasonably comprehensive tests that this code works as intended - since there's quite lot of logic going on here.

models/Event.ts
```diff
@@ -41,11 +41,11 @@ export interface QueryOptions {
 export interface EventStatic extends ModelStaticCommon<Event> {
   query: (
     user: User,
-    startTime: string,
-    endTime: string,
-    deviceId: DeviceId,
-    offset: number,
-    limit: number,
+    startTime: string | null,
```
hardiesoft, line 44: I would have expected the actual type if the variable is not defined to be `undefined`, rather than `null`

PR #322, opened 2020-06-08
Improve docker

clare:
Aim: Create a simple docker image for development and unit test purposes that will run on all environments in a timely fashion

- Use a custom docker image with nodejs, our dependencies, postgis and minio
- Added a docker ignore file to speed up building image
- Move migration utils (JS) into the migrations utils (and make it js again so we don't have to run the ts compiler to user it). 
- Runs npm install before copying code so it shouldn't need to run this every time. 

Thanks so much to @jackodsteel for his initial work improving docker (https://github.com/TheCacophonyProject/cacophony-api/pull/233) that I used as a base for this pull request.


hardiesoft:
When running via `npm run dev` I got the error ```ERROR: for server  Cannot create container for service server: Conflict. The container name "/cacophony-api" is already in use by container "5d6b25cc6
eed77ed674d2044a2a8d865ced59d294da549fc9a8bb30fe76cd64e". You have to remove (or rename) that container to be able to reuse that name.```

I can fix this myself, but I'm pretty sure the old `run.py` script handled that case for us?

hardiesoft:
Other than my above comment, it's looking great.  Starts up so much faster now!  Really nice work.

clare:
> When running via `npm run dev` I got the error `ERROR: for server Cannot create container for service server: Conflict. The container name "/cacophony-api" is already in use by container "5d6b25cc6 eed77ed674d2044a2a8d865ced59d294da549fc9a8bb30fe76cd64e". You have to remove (or rename) that container to be able to reuse that name.`
> 
> I can fix this myself, but I'm pretty sure the old `run.py` script handled that case for us?

@hardiesoft I think this is just a transition problem from the old container to the new one because I reused the name.  Once you have fixed it once you won't have to do it again.   I could change the container name to fix this.

hardiesoft:
That makes sense, I wouldn't worry in that case.

PR #327, opened 2020-07-23
Allow videos to be uploaded with metadata

gferraro:
Create tracks for a recording if they are supplied with the raw video


api/auth.ts
```diff
@@ -97,7 +97,7 @@ type AuthenticateMiddleware = (
  * Authenticate a JWT in the 'Authorization' header of the given type
  */
 const authenticate = function (
-  type: string | null,
+  types: string[] | null,
```
clare, line 100: This could potentially break other peoples code doing anything with our api.  What is the reason for this change?

test/test_recordings_filter.py
```diff
@@ -37,7 +37,7 @@ def test_latitude_longitude_filter(self, helper):
         bobsDevice = helper.given_new_device(self, "bobs_device", bobsGroup)
         rec = bobsDevice.upload_recording({"location": [20, 20]})
 
-        print("  The recording shoudl have a defautl precision of 100m")
+        print("  The recording shoudl have a default precision of 100m")
```
clare, line 40: missed shoudl <he he>

api/auth.ts
gferraro: This isn't called anywhere direclty outside of this class, and i have made all the authentication methods work with this change

PR #335, opened 2020-12-07
Super user ops

hardiesoft:
Adds two new api methods that can only be called by super-users, to support the ability to switch to viewing browse as if you are a different user on the system.

- listUsers
- adminAuthenicateAsUser

### DIFF
--- a/api/V1/AuthenticateUser.ts
+++ b/api/V1/AuthenticateUser.ts
@@ -95,12 +95,7 @@ export default function(app: Application) {
     [
       auth.authenticateAdmin,
       oneOf(
-        [
-          middleware.getUserByName(body),
-          middleware.getUserByName(body, "nameOrEmail"),
-          middleware.getUserByEmail(body),
-          middleware.getUserByEmail(body, "nameOrEmail")
-        ],
+        [middleware.getUserByName(body, "name")],
         "could not find a user with the given username or email"
       )
     ],
--- a/api/V1/AuthenticateUser.ts
+++ b/api/V1/AuthenticateUser.ts
@@ -77,26 +77,23 @@ export default function(app: Application) {
   );
 
   /**
-   * @api {post} /authenticate_other_user_as_admin/ Authenticate as any user if you are a super-user.
-   * @apiName AuthenticateOtherUserAsAdmin
+   * @api {post} /admin_authenticate_as_other_user/ Authenticate as any user if you are a super-user.
+   * @apiName AdminAuthenticateAsOtherUser
    * @apiGroup Authentication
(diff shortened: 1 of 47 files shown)


api/V1/AuthenticateUser.ts
```diff
@@ -76,6 +76,38 @@ export default function (app: Application) {
...
+   *
+   * @apiParam {String} name Username identifying a valid user account
+   *
+   * @apiSuccess {String} token JWT string to provide to further API requests
+   */
+  app.post(
+    "/admin_authenticate_as_other_user",
+    [
+      auth.authenticateAdmin,
+      oneOf(
```
clare, line 94: OneOf?   Is this the only way to give a useful error message?

test/test_user.py
```diff
@@ -97,3 +96,31 @@ def test_patch_username_password_length_requirements(self, helper):
         user.update(username=long_username, password=long_password)
         print("  Then when I then try to login using the newly changed data, it should also be successful")
         helper.login_with_username_password(long_username, long_password)
+
+    def test_admin_list_users(self, helper):
+        print("A super admin user can list all other users")
```
clare, line 101: Probably a better test is to create a username and check you can actually see user from admin.

test/test_user.py
```diff
@@ -97,3 +96,31 @@ def test_patch_username_password_length_requirements(self, helper):
...
+
+        print("A regular user can't list all other users")
+        non_admin_user = helper.given_new_user(self, "non_admin")
+        with pytest.raises(AuthorizationError):
+            print("When a regular user tries to list other users it should return an error")
+            non_admin_user.list_users()
+
+    def test_admin_user_viewing_as_other_user(self, helper):
+        admin_user = helper.admin_user()
+        users = admin_user.list_users()
```
clare, line 114: I would definitely create a user here.   Then there should be no need to call list_users which would mean that the test_admin user test is should succeed even if the list_users api stops working.   Making it more robust.

test/test_user.py
```diff
@@ -97,3 +96,31 @@ def test_patch_username_password_length_requirements(self, helper):
...
+        other_user = [user for user in users.json()["usersList"] if user['id'] == 2][0]['username']
+        assert other_user
+        print("Admin can login as another user using just their username")
+        other_user_creds = admin_user.admin_login_as_other_user(other_user)
+        assert other_user_creds
+
```
clare, line 120: Do we need to prove they actually see different results than the admin user?   Ie that the other_user_cred works? Or is that going too far?

clare (approved):
Looking good Jon.   This is going to be great!

PR #338, opened 2020-12-08
Fix Visit Grouping Incorrectly

gferraro:
Once a device had seen a animal it was always using the first visit (most recent) of this animal to group prior visits too. While it should of been using the last / current visit.
Reverse() function was sorting in place causing the list to be in reverse order.

Visits and Events should be in descending startTime order, so earliest incomplete and mark completed should be also done in reverse


hardiesoft (approved):
I'm going to take your word that this fixes the issues, since I can't see any easy way of verifying it in the UI :)

PR #341, opened 2020-12-15
use master tag in visits calculation

gferraro:
Return groupid and deviceid for processing so we can see the orignating device in cacophony processing


clare (approved):
Yay!

PR #343, opened 2021-01-21
One visit

gferraro:
Calculate a overriding a tag for all tracks in a recording.
This is done by tallying up the tags, and the tag with the highest count is assumed to be correct for all tracks.

Added assumedTag property to a VisitEvent, this refers to the overriding tag that this event has been given. 
E.g. it it was tagged as a cat but has been grouped under a possum visit, the assumedTag would be "possum" while tag would be "cat"

### DIFF
--- a/api/V1/Visits.ts
+++ b/api/V1/Visits.ts
@@ -26,24 +26,32 @@ interface AnimalMap {
   [key: string]: VisitSummary;
 }
 
-function getRecordingTag(tracks:any [], userID: number): any | null{
-  animalVote = {}
-  let max = null;
-  for(const track of tracks){
-    const tag = getTrackTag(track.TrackTags, userId)
-    if (tag && tag.what != unidentified) {
-        if(tag in animalVote){
-          animalVote[tag.what] +=1
-        }else{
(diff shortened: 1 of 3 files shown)


api/V1/Visits.ts
```diff
@@ -26,6 +26,35 @@ interface AnimalMap {
...
+  let maxVote = null;
+  for (const track of tracks) {
+    const tag = getTrackTag(track.TrackTags, userID);
+    if (tag) {
+      if (tag.what in animalVote) {
+        animalVote[tag.what].count += 1;
+      } else {
+        animalVote[tag.what] = { tag: tag, count: 1 };
+      }
+      if (
```
hardiesoft, line 42: This would probably be cleaner/clearer if you did the count accumulation as one step, and then just did an ascending sort of `Object.values(animalVote)` on the `count` key and popped off the last value as your result.  It's a bit finicky tracking this inside the accumulation loop IMO.

hardiesoft (approved):
Just a small suggestion, see comment, otherwise looks good, cheers

PR #345, opened 2021-02-02
updated visits algorithm

gferraro:
As per discussion with Grant.
A visit is now all tracks that occur within a specified interval (10 minutes).
All tracks of a single recording belong to the same visit.
A visit is tagged based on the tag which occurs the most and is not unidentified, unless no other tag is available.

Removed visits being specified in the animal map as well as in visits array. Instead just summarise the info per device and per animal


clare (approved):
Yay.   It looks a wee bit simpler now which should be good.

PR #347, opened 2021-02-09
Stations 2

hardiesoft:
Adds ability to manage stations


models/Group.ts
```diff
@@ -126,6 +142,165 @@ export default function (sequelize, DataTypes): GroupStatic {
...
+        return {
+          name: (s as Station).name,
+          location: (s as Station).location.coordinates
+        }
+      }
+    });
+    for (const a of allStations) {
+      for (const b of allStations) {
+        if (a !== b && a.name !== b.name) {
+          if (latLngApproxDistance(a.location, b.location) < MAX_STATION_DISTANCE_THRESHOLD_METERS) {
```
clare, line 221: would MIN_STATION_SEPARATION be a better name for MAX_STATION_DISTANCE_THRESHOLD_METERS

models/Group.ts
```diff
@@ -126,6 +142,165 @@ export default function (sequelize, DataTypes): GroupStatic {
...
+            Math.abs(stationToUpdate.location.coordinates[0] - station.lat) < EPSILON ||
+            Math.abs(stationToUpdate.location.coordinates[1] - station.lat) < EPSILON
+        ) {
+          // NOTE - Casting this as "any" because station.location has a special setter function
+          (stationToUpdate as any).location = [station.lat, station.lng];
+          stationToUpdate.lastUpdatedById = authUser.id;
+          addedStations.push(stationToUpdate);
+          stationOpsPromises.push(stationToUpdate.save());
+        } else if (applyToRecordingsFromDate) {
+          addedStations.push(stationToUpdate);
```
clare, line 261: Why - if it hasn't moved then you haven't added it.

models/Group.ts
```diff
@@ -126,6 +142,165 @@ export default function (sequelize, DataTypes): GroupStatic {
...
+            // Group id, and after date
+            GroupId: group.id,
+            createdAt: {
+              [Op.gte]: applyToRecordingsFromDate.toISOString()
+            }
+          }
+      );
+      const recordingsFromStartDate: Recording[] = await models.Recording.findAll(builder.get());
+      const recordingOpPromises = [];
+      for (const recording of recordingsFromStartDate) {
```
clare, line 291: But it might be in station without being the closest station?

models/Group.ts
```diff
@@ -126,6 +142,165 @@ export default function (sequelize, DataTypes): GroupStatic {
...
+
+      await Promise.all([...stationOpsPromises, ...retirePromises]);
+      // Now addedStations are properly resolved with ids:
+      // Now we can look for all recordings in the group back to startDate, and check if any of them
+      // should be assigned to any of our stations.
+
+      // Find matching recordings to apply stations to from `applyToRecordingsFromDate`
+
+      // Get recordings for group starting at date:
+      const builder = await new models.Recording.queryBuilder().init(
```
clare, line 279: Surely updateExistingRecordings() is another function.   I believe it should work in exactly the same way as when a recording comes in.  So instead of using the calculated version of addedStations (including all stations in this case) it should just go to the stations table - get all the stations and go from there.

api/V1/recordingUtil.ts
```diff
@@ -63,6 +61,45 @@ export interface RecordingQuery {
...
+  const lat1 = (a[0] * Math.PI) / 180;
+  const costLat1 = Math.cos(lat1);
+  const sinLat1 = Math.sin(lat1);
+  const lat2 = (b[0] * Math.PI) / 180;
+  const deltaLng = ((b[1] - a[1]) * Math.PI) / 180;
+  const part1 = Math.acos(sinLat1 * Math.sin(lat2) + costLat1 * Math.cos(lat2) * Math.cos(deltaLng));
+  return part1 * R;
+}
+
+async function tryToMatchRecordingToStation(recording: Recording): Promise<StationId | null> {
```
clare, line 82: Can't we use the database to do this or are you trying to get more accurate?

api/V1/recordingUtil.ts
```diff
@@ -63,6 +61,45 @@ export interface RecordingQuery {
   filterOptions: null | any;
 }
 
+// How close is a station allowed to be to another station?
+export const MAX_STATION_DISTANCE_THRESHOLD_METERS = 60;
+// The radius of the station is half the max distance between stations: any recording inside the radius can
```
clare, line 66: Yes but we aren't using this assumption are we?  We are currently doing a less then and then a sort.

models/Recording.ts
```diff
@@ -737,6 +752,26 @@ from (
   // INSTANCE METHODS
   //------------------
 
+  Recording.prototype.isInStation = function (station: Station | CreateStationData) {
+    if (station.hasOwnProperty('lat')) {
```
clare, line 756: Do we need this or should we always calculated for all stations and take minimum?

test/test_stations.py
```diff
@@ -0,0 +1,239 @@
...
+        assert len(retired) == 1
+        assert retired[0]["retiredAt"] is not None
+
+        # Get stations and make sure that "Will be retired" exists and is retired (getting group by name)
+        current_stations = station_user.get_stations_for_group(station_group_name)
+        retired = [x for x in current_stations["stations"] if x["name"] == "Will be retired"]
+        assert len(retired) == 1
+        assert retired[0]["retiredAt"] is not None
+
+        print("Adding stations with updated lat/lng but same name should update the lat/lng")
```
clare, line 107: I'm not sure we should be changing locations rather than storing a new one.   But it isn't the simplest solution for now so it's ok.

clare (changes requested):
This is a great start to a really really complicated piece of code.   

There are a few things to look at here. 

Also we need to add recalculating the station if the file processing adds the position to the recording record (as talked about with Cam). 

Thanks so much.

api/V1/Group.ts
```diff
@@ -163,4 +167,104 @@ export default function (app: Application, baseUrl: string) {
...
+  *
+  * @apiUse V1ResponseSuccess
+  * @apiUse V1ResponseError
+  */
+  app.post(
+    `${apiUrl}/:groupIdOrName/stations`,
+    [
+      auth.authenticateUser,
+      middleware.getGroupByNameOrIdDynamic(param, "groupIdOrName"),
+      body("stations")
```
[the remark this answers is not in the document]
hardiesoft: Can you elaborate on what you mean by this?

api/V1/recordingUtil.ts
hardiesoft: I think a database query made sense when we were wanting to do "live" matching of recordings to stations, but since we're just baking the results in when we add or update a recording, it's simpler to do it in client code.  This is the approach used by leaflet.js that I'm using for distance between two lat/lng points.

test/test_stations.py
```diff
@@ -0,0 +1,239 @@
...
+            station_user.add_stations_to_group(
+                station_group["groupId"],
+                json.dumps(
+                    [
+                        {"name": "Station name", "lat": -43.62367659982135, "lng": 172.62646754804894},
+                        {"name": "Too close station", "lat": -43.62367657911904, "lng": 172.62626365029485},
+                    ]
+                ),
+            )
+        print("Error", json.loads(str(error.value)))
```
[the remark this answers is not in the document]
hardiesoft: Thanks again for the awesome feedback.  I've addressed most issues, and will mostly leave the beautification/breaking up of the tests for when we are redoing them in cypress.

hardiesoft:
I think this has all been addressed now.

api/V1/Group.ts
```diff
@@ -191,26 +213,7 @@ export default function (app: Application, baseUrl: string) {
...
-                    lat: { type: "number" },
-                    lng: { type: "number" },
-                },
-                required: ["name", "lat", "lng"]
-            }
-        }, {throwFirst: true});
-        req.body.stations = stations;
-        return true;
-      }),
+          .custom(validateStationsJson),
```
clare, line 216: nice :)

models/Group.ts
```diff
@@ -185,120 +263,55 @@ export default function (sequelize, DataTypes): GroupStatic {
...
-        for (const station of addedStations) {
-          // Find any matches for stations
-          if (recording.isInStation(station)) {
-            recordingOpPromises.push(recording.setStation(station));
-            break;
-          }
-        }
-      }
-      return Promise.all(recordingOpPromises).then(() => addedStations.map(({id}) => id));
+      await Promise.all([...stationOpsPromises, ...retiredStations]);
```
clare, line 310: The change the stations has already been done hasn't it?

api/V1/recordingUtil.ts
```diff
@@ -79,23 +79,29 @@ export function latLngApproxDistance(a: [number, number], b: [number, number]):
   return part1 * R;
 }
 
-async function tryToMatchRecordingToStation(recording: Recording): Promise<StationId | null> {
+export async function tryToMatchRecordingToStation(recording: Recording, stations?: Station[]): Promise<Station | null> {
   // Match the recording to any stations that the group might have:
-  const group = await models.Group.getFromId(recording.GroupId);
-  const stations = await group.getStations();
+  if (!stations) {
```
clare, line 85: Recording when first uploaded may not have a GPS position I think.   If so we should just return straight away?

clare (approved):
Looking good.   Have a look at my comments though.

PR #348, opened 2021-02-17
Added Cypress tests

clare:
First look at the cypress tests.  Note that they are not running on travis yet!


hardiesoft:
Can we add prettier to our eslint config, and make sure it gets run.  Looks like this may not be happening properly/consistently for the actual api repo either.  Do we need to update eslint to keep the versions consistent?  Or should the eslint config for the tests not actually be here - should it be added to the scope of the outer project somehow?  Either way, something like `https://github.com/prettier/eslint-config-prettier` looks like what we'd want.

test-cypress/cypress/commands/api/camera.d.ts
```diff
@@ -0,0 +1,16 @@
+// load the global Cypress types
+/// <reference types="cypress" />
+
+declare namespace Cypress {
+    interface Chainable {
+        /**
+         * create a group for the given user (who has already been referenced in the test
+        */
+        apiCreateCamera(cameraname: string, groupname: String, log?: boolean): Chainable<Element>
```
hardiesoft, line 9: `String` should be `string` - there are a few instances of this.

hardiesoft, line 9: Can we be consistent with our lowerCamelCase naming scheme?  `cameraname` -> `cameraName`, `groupname` -> `groupName` etc

test-cypress/cypress/commands/api/recording.ts
```diff
@@ -0,0 +1,132 @@
...
+function addTracksToRecording(data : thermalRecordingData, model: string, trackDetails?: trackInfo[]) : void {
+  data.metadata = {
+    "algorithm": {
+      "model_name": model
+     },
+    "tracks": []
+  }
+
+  if (trackDetails) {
+    trackDetails.forEach((track) => {
```
hardiesoft, line 111: This would be better as `data.metadata.tracks = trackDetails.map((track) => {...})`

test-cypress/cypress/commands/api/camera.ts
```diff
@@ -0,0 +1,44 @@
...
+import {logTestDescription} from "../descriptions"; 
+
+Cypress.Commands.add("apiCreateCamera", (cameraName: string, group: string, log = true) => {
+  logTestDescription(`Create camera '${cameraName}' in group '${group}'`, {
+    camera: cameraName,
+    group: group
+  }, 
+  log);
+
+  const request = createCameraDetails(cameraName, group)
```
hardiesoft, line 12: Once prettier is installed it will fix the lack of semi-colons here ;-)

test-cypress/cypress/commands/server.ts
```diff
@@ -0,0 +1,85 @@
...
+   jwt: string,
+   id: number
+}
+
+export function getCreds(username: string) : apiCreds {
+   return Cypress.env('testCreds')[username];
+}
+
+export function saveCreds(response: Cypress.Response, name: string, id = 0) {
+   const creds = {
```
hardiesoft, line 32: In es6, you don't need to do `name: name`, `id: id` - you can just write { name, id, ... } etc.

test-cypress/cypress/commands/server.ts
```diff
@@ -0,0 +1,85 @@
...
+        'authorization': response.body.token
+      },
+      jwt: response.body.token,
+      id: id
+   }
+   Cypress.env('testCreds')[name] = creds;
+}
+
+export function checkRequestFails(requestdetails: any) {
+   requestdetails.failOnStatusCode = false;
```
hardiesoft, line 44: I don't understand why you add the `failOnStatusCode` property here - nothing seems to use it?

test-cypress/cypress/commands/server.ts
```diff
@@ -0,0 +1,85 @@
...
+      headers: {
+        'authorization': response.body.token
+      },
+      jwt: response.body.token,
+      id: id
+   }
+   Cypress.env('testCreds')[name] = creds;
+}
+
+export function checkRequestFails(requestdetails: any) {
```
hardiesoft, line 43: This only seems to be used with one shape of object, so can we be more specific than `any`?

test-cypress/cypress/commands/server.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export function v1ApiPath(page: string, queryParams : any = {}) : string {
+   const urlpage = url.format({
+      pathname:`/api/v1/${page}`,
+      query: queryParams
+    });
+   return `${Cypress.env('cacophony-api-server')}${urlpage }`;
+}
+
+interface apiCreds {
```
hardiesoft, line 18: Custom types should really be UpperCamelCase by convention

test-cypress/cypress/commands/server.ts
```diff
@@ -0,0 +1,85 @@
+// load the global Cypress types
+/// <reference types="cypress" />
+export const DEFAULT_DATE = new Date(2021, 1, 16, 22);
+import url = require("url");
```
hardiesoft, line 4: Mixing `require` with `import` seems weird - shouldn't this be `import url from 'url'`, or `import * as url from 'url'` if url has no default export?

test-cypress/cypress/commands/api/recording.d.ts
```diff
@@ -0,0 +1,30 @@
+// load the global Cypress types
+/// <reference types="cypress" />
+
+interface trackInfo {
```
hardiesoft, line 4: `TrackInfo` - plus are these fields supposed to be commented out?

test-cypress/cypress/commands/api/recording.ts
```diff
@@ -0,0 +1,132 @@
...
+  comment? : string,
+  batteryLevel? : number, 
+  batteryCharging? : string, 
+  airplaneModeOn? : boolean, 
+  version?: string, 
+  additionalMetadata?: JSON,
+  metadata?: thermalRecordingMetaData
+}
+
+function makeRecordingDataFromDetails(details: thermalRecordingInfo) : any {
```
hardiesoft, line 55: Aren't we always returning a `ThermalRecordingData` here?

test-cypress/cypress/commands/api/recording.ts
```diff
@@ -0,0 +1,132 @@
...
+}
+
+interface thermalRecordingMetaData {
+  algorithm? : algorithmMetadata,
+  tracks : trackData[]
+}
+
+interface thermalRecordingData {
+  type: "thermalRaw",
+  recordingDateTime: string,
```
hardiesoft, line 44: This is totally an optional style thing, but when I have a string in a particular format, or a number that I know represents an id of a particular type, I like to create a type alias for it.  Eg: `type IsoFormattedDateString = string`, or `type UserId = number` - and then use those aliases in interfaces to provide additional information about what is expected.

test-cypress/cypress/commands/api/recording.ts
```diff
@@ -0,0 +1,132 @@
...
+  tracks : trackData[]
+}
+
+interface thermalRecordingData {
+  type: "thermalRaw",
+  recordingDateTime: string, 
+  duration: number, 
+  comment? : string,
+  batteryLevel? : number, 
+  batteryCharging? : string,
```
hardiesoft, line 48: We can do better with types like this: see `Recording.ts:210` - there are a known set of allowed strings for `batteryCharging`.  If we're reusing that set often, we could declare it as an enum.

test-cypress/cypress/commands/api/user.ts
```diff
@@ -0,0 +1,115 @@
+// load the global Cypress types
+/// <reference types="cypress" />
+
+import url = require("url");
+import names = require("../names");
```
hardiesoft, line 5: `import { getTestName } from "../names" ?

test-cypress/cypress/commands/server.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export function apiPath() : string {
```
hardiesoft, line 6: Does this need to be an exported function?  Can it just be `export const API_PATH = Cypress.env('cacophony-api-server')` ?  I don't think it changes, and I assume that the env var is available at import time?

test-cypress/cypress/commands/api/user.ts
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  cy.request({
+    method: 'POST', 
+    url: groupURL, 
+    headers: user.headers,
+    body: data});
+});
+
+Cypress.Commands.add("apiCheckUserCanSeeGroup", (username, group) => {
```
hardiesoft, line 58: Seems to be missing from the `user.d.ts` file.

test-cypress/cypress/commands/api/user.d.ts
```diff
@@ -0,0 +1,26 @@
+// load the global Cypress types
+/// <reference types="cypress" />
+
+declare namespace Cypress {
+    interface Chainable {
+        /**
+         * create user and save api credentials further use
+        */
+        apiCreateUser(username: string, log: boolean): Chainable<Element>
```
hardiesoft, line 9: All of these custom commands are declared as returning `Chainable<Element>` - which I think is the correct convention for Cypress custome commands - but your actual commands don't return the cy.request which would actually make it chainable.

test-cypress/cypress/commands/server.ts
clare: Thanks for the tip.   Soo nice :)

test-cypress/cypress/commands/server.ts
clare: I've added a message.   Cypress uses it and will fail the test otherwise.

hardiesoft: Ah, so `any` should probably have been `Partial<Cypress.RequestOptions>` which is what you are passing to cypress

test-cypress/cypress/commands/api/recording.ts
clare: Like

hardiesoft (approved):

PR #350, opened 2021-02-22
Add processing fields

gferraro:
recordingDateTime and location, will be usefull for dennis and morepork detection

### DIFF
--- a/models/Recording.ts
+++ b/models/Recording.ts
@@ -1413,7 +1413,9 @@ from (
     "processingMeta",
     "GroupId",
     "DeviceId",
-    "StationId"
+    "StationId",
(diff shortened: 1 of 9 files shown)


models/Recording.ts
```diff
@@ -1413,7 +1414,9 @@ from (
     "processingMeta",
     "GroupId",
     "DeviceId",
-    "StationId"
+    "StationId",
+    "recordingDateTime",
+    "localtion"
```
hardiesoft, line 1419: Pretty sure this is a typo `localtion`

hardiesoft (changes requested):

hardiesoft (approved):

PR #355, opened 2021-03-01
Only show active devices

hardiesoft:
- Some changes to only return active devices from deviceListing API by default
- Warn, don't fail when adding stations that are close together
- Add simple APIs for getting group users, group devices, plus auth methods to check user permissions.


Dockerfile
```diff
@@ -4,7 +4,7 @@
 FROM cacophonyproject/server-base
 
 # NOTE: Using locally for arm64
-# FROM cacophony-api
+#FROM cacophony-api
```
clare, line 7: better than before?

api/V1/Group.ts
```diff
@@ -120,6 +121,47 @@ export default function (app: Application, baseUrl: string) {
     })
   );
 
+    app.get(
```
clare: These two new apis should probably have some *apidocs comments associated with them.

api/V1/Group.ts
```diff
@@ -252,25 +295,16 @@ export default function (app: Application, baseUrl: string) {
...
-      middleware.getGroupByNameOrIdDynamic(param, "groupIdOrName")
+      middleware.getGroupByNameOrIdDynamic(param, "groupIdOrName"),
+      auth.userHasReadAccessToGroup
     ],
     middleware.requestWrapper(async (request, response) => {
-      if (
-        request.user.hasGlobalRead() ||
-        (await request.user.isInGroup(request.body.group.id))
-      ) {
         const stations = await request.body.group.getStations();
```
clare, line 330: This is much neater....Yay!

docker-base/Dockerfile
```diff
@@ -23,14 +23,16 @@ RUN echo "host all all ::/0 md5" >> /etc/postgresql/10/main/pg_hba.conf
 RUN curl --location --fail --silent --show-error --remote-name https://dl.minio.io/server/minio/release/linux-amd64/minio
 
 # Note: uncomment if you're developing on arm64
-# RUN curl --location --fail --silent --show-error --remote-name https://dl.minio.io/server/minio/release/linux-arm64/minio
+#RUN curl --location --fail --silent --show-error --remote-name https://dl.minio.io/server/minio/release/linux-arm64/minio
```
clare, line 26: Also a less pretty change but fine if it was "prettified!"

models/Group.ts
```diff
@@ -61,7 +68,7 @@ const stationLocationHasChanged = (
 
 const checkThatStationsAreNotTooCloseTogether = (
```
clare, line 69: Should we pull all the stations stuff out into it's own file?   It doesn't feel like it really belongs in Groups.

models/Group.ts
```diff
@@ -106,10 +135,12 @@ const updateExistingRecordingsForGroupWithMatchingStationsFromDate = async (
   const builder = await new AllModels.Recording.queryBuilder().init(authUser, {
     // Group id, and after date
     GroupId: group.id,
-    createdAt: {
+    recordingDateTime: {
```
clare, line 138: Opps yes this is correct!  Great find.

models/Group.ts
```diff
@@ -262,13 +299,11 @@ export default function (sequelize, DataTypes): GroupStatic {
     group,
     stationsToAdd,
     applyToRecordingsFromDate
-  ): Promise<StationId[]> {
-    if (!(await group.userPermissions(authUser)).canAddStations) {
-      throw new AuthorizationError(
```
clare, line 267: Are you still checking user authorization here?  Sorry a bit hard to tell.

models/Group.ts
```diff
@@ -337,17 +372,40 @@ export default function (sequelize, DataTypes): GroupStatic {
...
         authUser,
         group,
         applyToRecordingsFromDate,
         allStations
       );
-      await Promise.all(updatedRecordings);
+      updatedRecordings = await Promise.all(updatedRecordings);
+    }
+    const result: {
+      stationIdsAddedOrUpdated: StationId[],
```
clare, line 387: Oooo nice!

clare (approved):
Yay great changes!

models/Group.ts
hardiesoft: Yep, just moved it up to the api middleware layer, since it seems more appropriate there.

models/Group.ts
hardiesoft: Yeah, I think next time I make changes I might do that - probably can just live at the top of Stations model, since I don't think we really want to create "utility function" files for each model at this stage.

models/User.ts
```diff
@@ -300,6 +302,13 @@ export default function (
     return groupIds.includes(groupId) === true;
   };
 
+  User.prototype.isGroupAdmin = async function (
+      groupId: number
+  ): Promise<boolean> {
+    const groupIds = await this.getGroups();
+    return groupIds.some(({id, GroupUsers}) => id === groupId && GroupUsers.admin === true);
```
gferraro, line 309: should this be `groupUser`?

gferraro (commented):
Looks good,  just a few little things

models/Group.ts
```diff
@@ -78,26 +85,48 @@ const checkThatStationsAreNotTooCloseTogether = (
       };
     }
   });
+  const tooClosePairs: Record<string, { station: NamedLocation, others: NamedLocation[] }> = {};
   for (const a of allStations) {
     for (const b of allStations) {
       if (a !== b && a.name !== b.name) {
```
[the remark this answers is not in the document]
hardiesoft: In this case, no, because the new stations to be created don't have ids yet.

models/User.ts
hardiesoft: Not according to sequelize I don't think.

models/Group.ts
```diff
@@ -106,10 +135,12 @@ const updateExistingRecordingsForGroupWithMatchingStationsFromDate = async (
   const builder = await new AllModels.Recording.queryBuilder().init(authUser, {
     // Group id, and after date
     GroupId: group.id,
-    createdAt: {
+    recordingDateTime: {
       [Op.gte]: fromDate.toISOString()
     }
   });
+  builder.query.distinct = true;
+  builder.query.limit = 10000000;
```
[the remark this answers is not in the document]
hardiesoft: Actually, probably the right thing to do is just `delete builder.query.limit`

models/Group.ts
```diff
@@ -78,26 +85,48 @@ const checkThatStationsAreNotTooCloseTogether = (
...
   for (const a of allStations) {
     for (const b of allStations) {
       if (a !== b && a.name !== b.name) {
         if (
           latLngApproxDistance(a.location, b.location) <
           MIN_STATION_SEPARATION_METERS
         ) {
-          throw new ClientError("Stations too close together");
+          if (!tooClosePairs.hasOwnProperty(a.name)) {
+            tooClosePairs[a.name] = { station: a, others: [] };
```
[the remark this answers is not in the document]
hardiesoft: Good call :)

PR #358, opened 2021-03-02
Cypress updates

clare:
New tests.


test-cypress/ARCHITECTURE.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Unique names for users, groups and cameras
+
+To make it possible to rerun tests without restarting the database, user, group and camera names are made unique across runs. 
+
+Thus the username "grant" would become something like cy_grant_010c03fb.  
+* **cy_**  is prepended to indicate that it is a test artifact.  (This allows us to delete such artifacts from our test server)
+* **_xxxxxxxx**  A random string is postpended to make it unique accross test runs.  The same random string is used for all tests within the same test run.
+
+Sometimes, when developing a test you might want to user the same user, group etc accross run.   You can do this by setting what the random string will be.   In the above case use:
```
hardiesoft, line 11: There is only one c in across :)

hardiesoft: `case use` should be `use case`?

test-cypress/ARCHITECTURE.md
```diff
@@ -0,0 +1,45 @@
...
+If you wish to log into the web app using an user account you created then the password is **p + username**.  For example in the above example the password would be **pcy_grant_010c03fb**
+
+
+## Running tests ##
+
+It goes without saying that all tests should run:
+  * Individually - as single it(....)
+  * As part of the suite of all tests
+
+You can reuse users, cameras etc within a file, but only if each of the individual tests don't rely object created by other parts of the test.
```
hardiesoft, line 28: rely *on" object(s)?

test-cypress/ARCHITECTURE.md
```diff
@@ -0,0 +1,45 @@
...
+
+To help future maintainers, all tests should have **logTestDescription** statements which describe what the test is doing.   These statements
+are logged in the development test runner as **TEST** which and should make it much easier for test developers to understand and therefore fix broken tests.   
+Most custom cy function should have a **logTestDescription** statement, and there may be other times that need them too. 
+
+When creating a **logTestDescription** you can also add a map of extra information.   This will show up in the console of the development test runner, 
+if the line is clicked.   So add any extra information that you think may be userful. 
+
+In the future we may also use these statements to describe what our tests do to document such behaviour as the criteria around visits.   In such areas the
+idea is quite simple, but there is enough complexity that only fine grain examples can really define what is happening.
```
hardiesoft, line 44: grain -> grained?

test-cypress/cypress/commands/api/recording.ts
```diff
@@ -1,29 +1,85 @@
...
+Cypress.Commands.add(
+  "uploadRecordingThenUserTag",
+  (
+    camera: string,
+    details: ThermalRecordingInfo,
+    tagger: string,
+    tag: string
+  ) => {
+    cy.uploadRecording(camera, details).then((inter) => {
+      cy.log(`request is ${JSON.stringify(inter.response.body)}`);
```
hardiesoft, line 80: A tip you may choose to use: JSON.stringify(obj, null, "\t") will format the stringified object with indenting, which may be nicer for users to read in the console?

test-cypress/cypress/commands/api/recording.ts
```diff
@@ -118,12 +175,19 @@ function addTracksToRecording(
     tracks: []
   };
 
+  if (tags && !trackDetails) {
+    trackDetails = tags.map((tag) => {
```
hardiesoft, line 179: An ES6 shorthand for this is `tags.map((tag) => ({tag}))`

test-cypress/cypress/commands/api/recording.ts
```diff
@@ -118,12 +175,19 @@ function addTracksToRecording(
...
+  }
+
   if (trackDetails) {
+    let count = 0;
     data.metadata.tracks = trackDetails.map((track) => {
       let tag = track.tag ? track.tag : "possum";
       return {
-        start_s: 2,
-        end_s: 8,
+        start_s: track.start_s || 2 + count * 10,
```
hardiesoft, line 189: It doesn't look like `count` ever gets incremented from zero?

test-cypress/cypress/commands/api/visits.ts
```diff
@@ -5,39 +5,117 @@ import { v1ApiPath, getCreds } from "../server";
...
+  user: string,
+  camera: string,
+  expectedVisits: ComparableVisit[]
+) {
+  const where: VisitsWhere = {
+    duration: { $gte: "0" },
+    type: "thermalRaw"
+  };
+
+  if (camera) {
```
hardiesoft, line 50: Can camera ever not be present?  That's not what the function signature says?

test-cypress/cypress/commands/api/visits.ts
```diff
@@ -5,39 +5,117 @@ import { v1ApiPath, getCreds } from "../server";
...
+    method: "GET",
+    url: v1ApiPath("recordings/visits", params),
+    headers: getCreds(user).headers
+  }).then((response) => {
+    checkResponseMatches(response, expectedVisits);
+  });
+}
+
+function checkResponseMatches(
+  response: any,
```
hardiesoft, line 69: `any` should be `cy.Response`?

test-cypress/cypress/commands/api/visits.ts
```diff
@@ -5,39 +5,117 @@ import { v1ApiPath, getCreds } from "../server";
...
+  expectedVisits: ComparableVisit[]
+) {
+  const responseVisits = response.body.visits;
+
+  expect(
+    responseVisits.length,
+    `Number of visits to be ${responseVisits.length}`
+  ).to.eq(expectedVisits.length);
+  const increasingDateResponseVisits = responseVisits.reverse();
```
hardiesoft, line 78: Just remember that `.reverse()` on an array in JS also mutates the original array in place - which may or may not be what you expect here.

test-cypress/cypress/commands/api/visits.ts
```diff
@@ -5,39 +5,117 @@ import { v1ApiPath, getCreds } from "../server";
...
+
+    if (expectedVisit.tag) {
+      simplifiedResponseVisit.tag = completeResponseVisit.what || "<null>";
+    }
+
+    if (expectedVisit.recordings) {
+      const recordingIds = completeResponseVisit.events.map((ev) => {
+        return ev.recID;
+      });
+      const uniqueRecordingIds = recordingIds.reduce((acc, recId) => {
```
hardiesoft, line 95: This isn't the best way to de-dupe an array.  Better would be to do `Object.keys(recordingIds.reduce((acc, recId) -> { acc[recId] = true; return acc;})) or just use the built-in Set object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set.

Totally a nit-pick though, and really is only for readability, this would only be inefficient for really big values of n.

test-cypress/cypress/commands/api/visits.ts
```diff
@@ -5,39 +5,117 @@ import { v1ApiPath, getCreds } from "../server";
...
+      simplifiedResponseVisit.start = completeResponseVisit.start;
+    }
+
+    if (expectedVisit.end) {
+      simplifiedResponseVisit.end = completeResponseVisit.end;
+    }
+
+    responseVisitsToCompare.push(simplifiedResponseVisit);
+  }
+  expect(JSON.stringify(responseVisitsToCompare)).to.eq(
```
hardiesoft, line 114: I'm not 100% sure if we can rely on two different objects that are the same having the same JSON.stringify representation - there seems to be variance across JS engines as to whether stringify is actually deterministic in terms of things like object key order.  It sounds like object key order might actually be dependent on the creation order of the object properties?  If you do 
```
let obj1 = {};
let obj2 = {};
obj1["a"] = true;
obj1["b"] = true;
obj2["b"] = true;
obj2["a"] = true;
```
and stringify both, the strings will not be the same.  This might cause things to fail sometimes?

test-cypress/cypress/commands/fileUpload.ts
```diff
@@ -0,0 +1,77 @@
...
+            fileName,
+            fileType,
+            uploader: credName,
+            data,
+            response: xhr.response
+          };
+        }
+      });
+
+      if (xhr.status != 200) {
```
hardiesoft, line 72: If the if statement around this necessary?

hardiesoft (approved):
Looks great, so much more understandable!  Just a few things you might want to check out before merging.

Thanks!

test-cypress/cypress/commands/api/visits.ts
clare: Yes.   These are tests so if we see a problem then we'll do it differently.  If the are arrays are different it will fail and that what's we need to happen.

test-cypress/cypress/commands/api/visits.ts
clare: I tried set but had ts lib set to es5.   All fixed now.   Yay!

test-cypress/cypress/commands/api/recording.ts
clare: Thanks for this tip!

PR #359, opened 2021-03-08
Make global permissions explicit

hardiesoft:
All api listing commands will automatically return all results if the user is a super-admin

user.  This PR adds an optional "view-mode" query param to these APIs,
which, if set to "limited", will restrict results to items only directly viewable by the super-admin user.
This is then plumbed through to the various database queries generated.

This does not change any existing behaviour, as that is still the default.


api/V1/Group.ts
```diff
@@ -121,6 +126,108 @@ export default function (app: Application, baseUrl: string) {
...
+    [
+        auth.authenticateUser,
+        middleware.getGroupByNameOrIdDynamic(param, "groupIdOrName"),
+        auth.userHasReadAccessToGroup
+    ],
+    middleware.requestWrapper(async (request, response) => {
+        const devices = await request.body.group.getDevices({ where: { active: true}, attributes: ["id", "devicename" ]});
+        return responseUtil.send(response, {
+            statusCode: 200,
+            data: devices.map(({id, devicename}) => ({
```
clare, line 187: Should we named our data eg:"devices"   It has been rows up until now.

api/V1/Group.ts
```diff
@@ -121,6 +126,108 @@ export default function (app: Application, baseUrl: string) {
     })
   );
 
+    /**
```
clare, line 129: Yay all the functions we needed :)

api/middleware.ts
```diff
@@ -215,6 +216,17 @@ const checkNewPassword = function (field: string): ValidationChain {
   });
 };
 
+const viewMode = function(): ValidationChain {
+  // All api listing commands will automatically return all results if the user is a super-admin
+  // There is now an optional "view-mode" query param to these APIs, which, if set to "limited",
+  // will restrict results to items only directly viewable by the super-admin user.
+  // The default behaviour remains unchanged, and this will do nothing for non-admin users.
+  return query("view-mode").custom((value, {req}) => {
+    req.body.viewAsSuperAdmin = value !== "limited";
```
clare, line 225: I find it a bit confusing that it is a query test but then you are setting the value in the body.   Mostly I am just finding it confusing because it is magic!

models/Recording.ts
```diff
@@ -101,7 +101,8 @@ interface RecordingQueryBuilder {
     tags?: string[], // AcceptableTag[]
     offset?: number,
     limit?: number,
-    order?: Order
+    order?: Order,
```
clare, line 104: Maybe we need an object here not a ridiculous list of params.

clare (approved):
Thanks for making this change.   I guess you might of missed a couple of things but hopefully we can sort these out later....

PR #360, opened 2021-03-11
Add alert for stopped devices

gferraro:
Add function to eventUtil to return the latest power events per device
Added cypress tests

### DIFF
--- a/api/V1/Event.ts
+++ b/api/V1/Event.ts
@@ -210,12 +210,9 @@ export default function (app: Application, baseUrl: string) {
    *       "lastReported": "2021-03-09 08:00:00",
    *       "lastStarted": "2021-03-09 18:00:00",
    *       "lastStopped": "2021-03-10 08:00:00",
-   *       "Device": {"id": 10, "devicename": "Camera2", "groupname": "examplegroup"},
-   *       "AdminUsers":
-   *       [
-   *        {"id":2, "email": admin@pest.co.nz, "firstname": John, "lastname": Doe, admin: True}
-   *       ]
+   *       "Device": {"id": 10, "devicename": "Camera2", "groupname": "examplegroup", "Group":{"id": 2}},
    *       "hasStopped": true
+   *       "hasAlerted": true
    *     }]
    * @apiUse V1ResponseError
    */
--- a/api/V1/Event.ts
+++ b/api/V1/Event.ts
@@ -211,6 +211,10 @@ export default function (app: Application, baseUrl: string) {
    *       "lastStarted": "2021-03-09 18:00:00",
    *       "lastStopped": "2021-03-10 08:00:00",
    *       "Device": {"id": 10, "devicename": "Camera2", "groupname": "examplegroup"},
+   *       "AdminUsers":
+   *       [
+   *        {"id":2, "email": admin@pest.co.nz, "firstname": John, "lastname": Doe, admin: True}
+   *       ]
    *       "hasStopped": true
    *     }]
    * @apiUse V1ResponseError
--- a/api/V1/Event.ts
+++ b/api/V1/Event.ts
@@ -195,39 +195,36 @@ export default function (app: Application, baseUrl: string) {
     })
   );
 
-    /**
-     * @api {get} /api/v1/events/powerEvents Query power events for devices
-     * @apiName QueryPower
-     * @apiGroup Events
-     *
-     * @apiUse V1UserAuthorizationHeader
-     * @apiParam {Integer} [deviceId] Return only errors for this device id
-     *
-     * @apiSuccess {JSON} array of device power events
-     * @apiSuccessExample {json} Success-Response:
-      *     HTTP/1.1 200 OK
-      *     "rows": [{
-      *       "lastReported": "2021-03-09 08:00:00",
-      *       "lastStarted": "2021-03-09 18:00:00",
-      *       "lastStopped": "2021-03-10 08:00:00",
-      *       "Device": {"id": 10, "devicename": "Camera2", "groupname": "examplegroup"},
-      *       "hasStopped": true
-      *     }]
-     * @apiUse V1ResponseError
-     */
-    app.get(
-      apiUrl + "/powerEvents",
-      [
-        auth.authenticateUser,
-        query("deviceId").isInt().optional().toInt(),
-      ],
-      middleware.requestWrapper(async (request, response) => {
-        const result = await eventUtil.powerEventsPerDevice(request);
-        return responseUtil.send(response, {
-          statusCode: 200,
-          messages: ["Completed query."],
-          rows: result
-        });
-      })
-    );
+  /**
+   * @api {get} /api/v1/events/powerEvents Query power events for devices
+   * @apiName QueryPower
+   * @apiGroup Events
+   *
+   * @apiUse V1UserAuthorizationHeader
+   * @apiParam {Integer} [deviceId] Return only errors for this device id
+   *
+   * @apiSuccess {JSON} array of device power events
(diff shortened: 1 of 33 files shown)


api/V1/eventUtil.ts
```diff
@@ -79,8 +112,80 @@ const eventAuth = [
...
+  }
+  checkIfStopped(): boolean {
+    if (this.lastStarted == null) {
+      this.hasStopped = false;
+      return this.hasStopped;
+    }
+    let hasStopped = false;
+
+    // check that started hasn't occured after stopped
+    if (
```
clare, line 158: Decouple this if statement, 

It should be:
if (laststopped == null) 
..
else if this.lastStarted.isAfter....

else

It makes no logical sense to combine the first two if statements then separate them.

clare (changes requested):
Looks nice and neat. Great work.   Just a few things and make sure it's going to the right people.

api/V1/Event.ts
```diff
@@ -194,4 +194,37 @@ export default function (app: Application, baseUrl: string) {
...
+   * @apiName QueryPower
+   * @apiGroup Events
+   *
+   * @apiUse V1UserAuthorizationHeader
+   * @apiParam {Integer} [deviceId] Return only errors for this device id
+   *
+   * @apiSuccess {JSON} array of device power events
+   * @apiSuccessExample {json} Success-Response:
+   *     HTTP/1.1 200 OK
+   *     "rows": [{
```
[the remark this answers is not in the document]
gferraro: I Agree i was just worried that there was some specific interface t hat was epecting "rows" in every response. But i changed it for alerts so will change that

api/V1/eventUtil.ts
```diff
@@ -79,8 +112,80 @@ const eventAuth = [
...
+    this.update(event);
+  }
+  update(event: Event) {
+    if (!this.Device && event.Device) {
+      this.Device = event.Device;
+    }
+    const eventDate = moment(event.dateTime);
+    switch (event.EventDetail.type) {
+      case "rpi-power-on":
+        if (this.lastStarted == null || eventDate.isAfter(this.lastStarted)) {
```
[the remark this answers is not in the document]
gferraro: That is true, but this class could be used with different queries

test-cypress/cypress/commands/api/user.d.ts
```diff
@@ -35,5 +35,13 @@ declare namespace Cypress {
      * create user group and camera at the same time
      */
     apiCreateUserGroup(userName: string, group: string): Chainable<Element>;
+
```
gferraro, line 38: Was using this but no longer need too, should we keep it anyway for future use?

api/V1/eventUtil.ts
```diff
@@ -69,22 +74,118 @@ async function uploadEvent(request, response) {
...
+
+    // check that started hasn't occured after stopped
+    if (this.lastStopped == null) {
+      // check that the started event was atleast 12 hours ago
+      this.hasStopped = moment().diff(this.lastStarted, "hours") > 12;
+    } else if (this.lastStarted.isAfter(this.lastStopped)) {
+      //check we are atleast 30 minutes after expected stopped time (based of yesterdays)
+      this.hasStopped = moment().diff(this.lastStopped, "minutes") > 24.5 * 60;
+    } else {
+      // check this isn't yesterdays start stop events
```
clare, line 171: Great this is much clearer.  Though this comment confused me.   Maybe "check that it started in the last 24 hours might be better".

api/V1/Event.ts
```diff
@@ -194,4 +194,38 @@ export default function (app: Application, baseUrl: string) {
...
+   * @apiParam {Integer} [deviceId] Return only errors for this device id
+   *
+   * @apiSuccess {JSON} array of device power events
+   * @apiSuccessExample {json} Success-Response:
+   *     HTTP/1.1 200 OK
+   *     "events": [{
+   *       "lastReported": "2021-03-09 08:00:00",
+   *       "lastStarted": "2021-03-09 18:00:00",
+   *       "lastStopped": "2021-03-10 08:00:00",
+   *       "Device": {"id": 10, "devicename": "Camera2", "GroupId": 22, "Group":{"groupname": "examplegroup"}},
```
clare, line 213: Can we flatten the groupname?  So it's device.groupname not device.group.groupname.  

This is something that Jon has pointed out we should be doing - keeping our apis cleaner. Lets have a skype call with the whole team about this. 

Even better would be to get the group name from the request of admins for the group because otherwise we are just repeating stuff.

models/Device.ts
```diff
@@ -45,8 +45,10 @@ export interface Device extends Sequelize.Model, ModelCommon<Device> {
     group: Group,
     newPassword: string
   ) => Promise<Device>;
-
+  Group: Group;
```
clare, line 48: Doesn't make sense to have group and group ID?

models/Event.ts
```diff
@@ -157,5 +167,61 @@ export default function (sequelize, DataTypes) {
...
+      }
+    }
+
+    let order: any[] = [
+      ["EventDetail", "type", "DESC"],
+      ["DeviceId", "DESC"],
+      ["dateTime", "DESC"]
+    ];
+
+    return this.findAndCountAll({
```
clare, line 195: I think you only need to do a findAll here not a findAllAndCount.  (one less request for the sql server).

models/Group.ts
```diff
@@ -153,32 +165,45 @@ const updateExistingRecordingsForGroupWithMatchingStationsFromDate = async (
...
 
 export interface Group extends Sequelize.Model, ModelCommon<Group> {
   id: GroupId;
+  groupname: string;
   addUser: (userToAdd: User, through: any) => Promise<void>;
   addStation: (stationToAdd: CreateStationData) => Promise<void>;
-  getUsers: (options?: { where?: any, attributes?: string[] }) => Promise<User[]>;
-  getDevices: (options?: { where?: any, attributes?: string[] }) => Promise<Device[]>;
+  getUsers: (options?: {
+    through?: any;
```
clare, line 188: through?

report-stopped-devices.ts
```diff
@@ -0,0 +1,139 @@
...
+import models from "./models";
+
+const ADMIN_EMAILS = ["support@2040.co.nz"];
+
+async function getUserEvents(powerEvents: PowerEvents[]) {
+  const groupAdmins = {};
+  const userEvents = {};
+
+  for (const event of powerEvents) {
+    if (!groupAdmins.hasOwnProperty(event.Device.GroupId)) {
```
clare, line 15: You are not even using group name here!

test-cypress/cypress/commands/api/user.d.ts
clare: If it works then we should.   As I've used it already in the other tests.

clare (approved):
Looks good GP.   We do need to have to have a bit of a chat about apis and how we should build them.   Please change countandFindAll to findall.

models/Device.ts
gferraro: Why? groupId belongs to Device table, if you fetch a Device with the associated Group it will be under the name Group. If you don't define it TS  throws a fit

api/V1/eventUtil.ts
gferraro: yep

models/Group.ts
gferraro: can be used to filter on the through table e.g. `through: { where: { admin: true } }`

report-stopped-devices.ts
gferraro: Should i be ....?

PR #361, opened 2021-03-15
Fixed inconsistency in the event-type name checking.

clare:
The event api allowed event types with '-' in them and many events types have these spaces in them.  However the get events request didn't allow anything other than alpha characters.   This anomaly how now been fixed.


hardiesoft (approved):
Nice!

PR #362, opened 2021-03-23
Visits update

gferraro:
Choose human tags over AI
If human tags dont agree consider the visit conflicting

### DIFF
--- a/api/V1/Visits.ts
+++ b/api/V1/Visits.ts
@@ -358,30 +358,44 @@ class Visit {
     this.addRecording(rec, userID);
   }
 
-  mostCommonTag(): string | null {
+  mostCommonTag(): [boolean| null, string | null] {
     // from all events in a visit, get the tag with the highest occurence that
     // isnt unidentified
     const tagCount = this.tagCount;
     const sortedKeys = Object.keys(tagCount).sort(function (a, b) {
-      if (a == unidentified) {
+      const [type_a, what_a] = a.split("-");
+      const [type_b, what_b] = b.split("-");
+      if(type_a != type_b){
+        // human tag takes precedence
+        if (type_a == "human"){
+          return -1;
+        }else{
+          return 1;
+        }
+      }
+      if (what_a == unidentified) {
         return 1;
-      } else if (b == unidentified) {
+      } else if (what_b == unidentified) {
         return -1;
       }
       return tagCount[b] - tagCount[a];
     });
+
     const maxVote = sortedKeys[0];
     if (maxVote) {
-      return maxVote;
+      const [type, what] = maxVote.split("-")
+      const humanTag = type == "human" ? true : false
+      return [humanTag, what];
     }
-    return null;
+    return [null,null];
   }
   completeVisit() {
     // assign the visit a tag based on the most common tag that isn't unidentified
-    const visitTag = this.mostCommonTag();
-    this.what = visitTag;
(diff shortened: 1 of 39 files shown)


api/V1/Visits.ts
```diff
@@ -18,7 +18,13 @@ import { Event } from "../../models/Event";
 let visitID = 1;
 const eventMaxTimeSeconds = 60 * 10;
 const aiName = "Master";
-const unidentified = "unidentified";
+const conflictTag = "conflicting tags";
+
+const metaTags = ["part", "poor tracking"];
+const unidentifiedTags = ["unidentified", "unknown"];
+const nonAnimalTags = [...metaTags];
```
clare, line 25: Can just use:
const nonAnimalTags = [...metaTags, ...unidentifiedTags]

api/V1/Visits.ts
```diff
@@ -50,19 +56,23 @@ function getTrackTag(trackTags: TrackTag[], userID: number): TrackTag | null {
...
-    const userTag = manualTags.find((tag) => tag.UserId == userID);
-    if (userTag) {
-      return userTag;
-    } else {
-      return manualTags[0];
+    const animalTags = manualTags.filter(
+      (tag) => !nonAnimalTags.includes(tag.what)
+    );
+    const uniqueTags = new Set(animalTags.map((tag) => tag.what));
+    if (uniqueTags.size > 1) {
```
clare, line 63: Do we want to return a conflict now, or after we analyse the whole visit?   We are losing data.   It is a rare suituation so I guess it doesn't matter much. Do we want to say what the conflict is between?

api/V1/Visits.ts
```diff
@@ -50,19 +56,23 @@ function getTrackTag(trackTags: TrackTag[], userID: number): TrackTag | null {
...
+      const conflict = {
+        what: conflictTag,
+        confidence: manualTags[0].confidence,
+        automatic: false
+      };
+      return conflict as TrackTag;
     }
+
+    return animalTags.length == 0 ? manualTags[0] : animalTags[0];
```
clare, line 72: Still don't want to return a meta tag or do you filter these out later.

api/V1/Visits.ts
```diff
@@ -358,30 +367,53 @@ class Visit {
...
-  mostCommonTag(): string | null {
+  mostCommonTag(): [boolean | null, string | null] {
     // from all events in a visit, get the tag with the highest occurence that
-    // isnt unidentified
+    // isnt unidentified, preferring human tags over ai
+    // returns [boolean desciribing if human tag, the tag]
     const tagCount = this.tagCount;
     const sortedKeys = Object.keys(tagCount).sort(function (a, b) {
-      if (a == unidentified) {
+      let hyphen = a.indexOf("-");
```
clare, line 376: This is all quite mysterious reading it.   Maybe it would be simpler if you define a type for the key, so you can use a.type and a.what.    Else make a function split to do the splitting.   You repeat the splitting three times in this one function.  

May just be easier to have a list of human tags and a list of ai tags run through the whole procedure.

clare (approved):
Good update. 

Thanks GP.

api/V1/Visits.ts
gferraro: whoops, thought i added that

api/V1/Visits.ts
gferraro: Good point have update to save the track tag

PR #366, opened 2021-04-12
Visits speed

gferraro:
Calculate audiobait events seperately to visits, then combine. This speeds up the request

Also set location before setting station id

### DIFF
--- a/api/V1/Visits.ts
+++ b/api/V1/Visits.ts
@@ -330,7 +330,7 @@ class DeviceVisits {
     return this.visits;
   }
 
-  addAudioBait(events: any) {
+  addAudioBait(events: Event[]) {
     for (const visit of this.visits) {
       visit.addAudioBaitEvents(events);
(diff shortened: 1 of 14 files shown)


api/V1/Visits.ts
```diff
@@ -416,31 +423,36 @@ class Visit {
...
   }
 
-  addAudioBaitEvents(rec) {
+  addAudioBaitEvents(allEvents: Event[]) {
     // add all audio bait events that occur within audioBaitInterval of this visit
     // and before the end of the visit
-    let events = rec.Device.Events;
-    if (!events) {
+    // let events = rec.Device.Events;
+    if (!allEvents) {
```
hardiesoft, line 431: You say in the function signature that the type is `Event[]`, but then you're checking to see if `allEvents` doesn't exist.  Should the type actually be something like `Event[] | undefined`?

hardiesoft (approved):
Looks good, thanks!

PR #367, opened 2021-04-14
Monitoring

clare:
Added a new monitoring api that allows us to do visits in a paged fashion.


api/V1/Monitoring.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * @api {get} /api/v1/monitoring/page Get visits page
+     * @apiName MonitoringPage
+     * @apiGroup Monitoring
+     * @apiDescription Get a page of monitoring visits.   First recording are sorted into visits and then the best-classification for this visit is calculated.
+     * Optionally you can also a specify an ai so you can compare the best classification with that given by the ai.   
+     * 
+     * @apiUse V1UserAuthorizationHeader
+     * @apiParam {number} devices  A list of ids of devices to include
```
hardiesoft, line 38: When you say a list of ids, what do you mean?  A JSON array of numbers?  The `@apiParam` hint just says `{number}`, which doesn't seem correct?

api/V1/Monitoring.ts
```diff
@@ -0,0 +1,150 @@
...
+     * @apiParam {number} groups  A list of ids of groups to include
+     * @apiParam {date} from  Retreive visits after this date
```
hardiesoft, line 40: Typo `Retreive` -> `Retrieve`

api/V1/Monitoring.ts
```diff
@@ -0,0 +1,150 @@
...
+     * @apiParam {date} until Retreive visits starting on or before this date
+     * @apiParam {number} page  Page number to retrieve
+     * @apiParam {number} page Maximum numbers of visits to show on each page
+     * @apiParam {string} ai   Name of AI to compute results to put in classificationAI.   This will not affect the (best) classification, 
+     * which always uses a predefined AI choice.
+     * @apiParam {string} viewmode   dsaf
```
hardiesoft, line 46: `dsaf`?

api/V1/monitoringPage.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (params.groups) {
+        criteria.groups = params.groups;
+    }
+
+    return criteria;
+}
+
+function replaceInSQL(sql: string, replacements: { [key: string]: string} ) : string {
+    for (const key in replacements) {
+        sql = sql.replace(`{${key}}`, replacements[key]);
```
hardiesoft, line 138: This seems suspect - key is a string, but you're using string interpolation on it by itself.  Why?
Also, string replace as used will only replace the first instance of the string found, which doesn't seem to be what you want.  You need to use a regex with the global flag set: `new RegExp(key, 'g')` : `https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace`

api/V1/monitoringPage.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    const countRet = await models.sequelize.query(replaceInSQL(VISITS_COUNT_SQL, replacements), { type: QueryTypes.SELECT });
+    const approxVisitCount = parseInt(countRet[0].count);
+    const returnVal = createPageCriteria(params, approxVisitCount);
+
+    if (approxVisitCount < params.pageSize) {
+        returnVal.pageFrom = returnVal.searchFrom;
+        returnVal.pageUntil = returnVal.searchUntil;        
+    } else if (params.page <= returnVal.pagesEstimate ) {
+        const limit : number = params.pageSize * 1 + 1;
```
hardiesoft, line 98: `pageSize * 1` seems redundant?

api/V1/Monitoring.ts
```diff
@@ -0,0 +1,150 @@
...
+     * @apiDescription Get a page of monitoring visits.   First recording are sorted into visits and then the best-classification for this visit is calculated.
+     * Optionally you can also a specify an ai so you can compare the best classification with that given by the ai.   
+     * 
+     * @apiUse V1UserAuthorizationHeader
+     * @apiParam {number} devices  A list of ids of devices to include
+     * @apiParam {number} groups  A list of ids of groups to include
+     * @apiParam {date} from  Retreive visits after this date
+     * @apiParam {date} until Retreive visits starting on or before this date
+     * @apiParam {number} page  Page number to retrieve
+     * @apiParam {number} page Maximum numbers of visits to show on each page
```
gw7nvw, line 43: Default value (100).  Max value (100)

api/V1/Monitoring.ts
gw7nvw, line 40: 'timestamp' or 'yyyy-mm-dd' date

api/V1/Monitoring.ts
gw7nvw, line 43: This has been implemented as 'page-size', shown here as 'page'. Both are inconsistent with return parameter naming convention which would have it as pageSize

clare: While this woud be ideal I am worried that it will get out of date as we change things.   It should error if the the user tries to retrieve more than 100.   I believe errrors are more useful documentation than the api docs.  As they are the information when you need it, where you need it.

hardiesoft (approved):
Looks good to me, merge away!

PR #372, opened 2021-04-29
Alert master

gferraro:
Only send alerts on master tag


hardiesoft (approved):
Cool, looks good to me

PR #375, opened 2021-05-06
check for null value, also add trackid to alert link

gferraro:
Previous change to remove EventDetail as a requirement from sequelize join (to speed up query) broke the if statement as previously checking just for event was enough to ensure eventdetail also existed.

### DIFF
--- a/emailUtil.ts
+++ b/emailUtil.ts
@@ -10,15 +10,13 @@ function alertBody(
   tag: TrackTag,
   camera: String
 ): string[] {
-  const dateTime = moment(recording.recordingDateTime)
-    .tz(config.timeZone)
-    .format("h:mma Do MMM");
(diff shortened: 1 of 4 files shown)


emailUtil.ts
```diff
@@ -10,15 +10,13 @@ function alertBody(
   tag: TrackTag,
   camera: String
 ): string[] {
-  const dateTime = moment(recording.recordingDateTime)
-    .tz(config.timeZone)
```
CameronRP, line 14: is setting the time zone not needed anymore?

CameronRP (approved):
Looks good

emailUtil.ts
gferraro: doh thanks

PR #376, opened 2021-05-09
Fix multi

gferraro:
4.2.1 seems to be calling close twice, causing errors.
Also chmod package-lock file no matter what to avoid install issues


hardiesoft (approved):
Sweet, thanks for tracking that down

PR #384, opened 2021-06-02
Alerts first

gferraro:
file processing will serve recordings for devices with alerts before others.
and changed fall back order to be by recordingdatetime asc then id ascending


hardiesoft (approved):
Looks good to me, cheers

PR #391, opened 2021-06-23
Updated monitoring paging to use duration and check time from end of previous recording not start.

clare:
I have analysed this code and it is not significantly slower than the old code.   Despite how the query is written the database knows to only build the Partition table once.


hardiesoft (approved):
Awesome, I look at pulling this into our test release today.

PR #393, opened 2021-06-27
Reprocess state

gferraro:
Add actual reprocess state, this will only work with cacophony-processing >= 2.8.5


api/fileProcessing/index.ts
```diff
@@ -99,8 +99,7 @@ export default function (app: Application) {
     }
 
     if (success) {
-      const jobs = models.Recording.processingStates[recording.type];
-      const nextJob = jobs[jobs.indexOf(recording.processingState) + 1];
+      const nextJob = recording.getNextState();
```
clare, line 102: Does this really work for reprocess?  Jobs.indexOf("reprocess") will return -1 won't it?

models/Recording.ts
```diff
@@ -773,6 +776,23 @@ from (
...
   //------------------
+  Recording.prototype.getNextState = function () {
+    const jobs = Recording.processingStates[this.type];
+    let nextState;
+    if (this.processingState == RecordingProcessingState.Reprocess) {
+      nextState = jobs[jobs.length - 1];
+    } else {
+      const job_index = jobs.indexOf(this.processingState);
+      if (job_index == -1) {
+        throw new Error(`Recording state unknown - ${this.processState}`);
```
clare, line 787: You need this error of job_index == -1 in your code otherwise it could do strange things.

clare (dismissed):
Still wonder if this really works for reprocess and it needs that error.   Then happy to have it tested.

api/fileProcessing/index.ts
gferraro: sorry i dont follow

hardiesoft (approved):

PR #395, opened 2021-07-07
Alert on finished

gferraro:
Only send alerts when a recording has come from analyse state and is finished
User the visit logic to get a tag for the whole recording and only alert on ai master tag of this
Also added cypress tests
Add alert in fileprocessing api if we upload videos into finished state (good for testing with)

tested by matt to be woroking


api/V1/recordingUtil.ts
```diff
@@ -1098,6 +1106,37 @@ function addAudioBaitRow(out: any, audioBait: Event) {
     ""
   ]);
 }
+
+async function sendAlerts(recID: number) {
+  const recording = (await models.Recording.getForAdmin(recID)) as any;
```
hardiesoft, line 1111: As discussed, not sure that having the `getForAdmin` abstraction is worthwhile until we have more consumers of that functionality - probably better to just inline the logic here.

api/V1/recordingUtil.ts
```diff
@@ -1098,6 +1106,37 @@ function addAudioBaitRow(out: any, audioBait: Event) {
...
+    if (matchedTag) {
+      matchedTrack = track;
+      break;
+    }
+  }
+  if (!matchedTag) {
+    return;
+  }
+
+  const alerts = await (models.Alert as AlertStatic).getActiveAlerts(
```
hardiesoft, line 1129: It seems like it would be much less work to check for active alerts at the top of this function, and exit early if there are no alerts?

test-cypress/cypress/commands/api/recording.d.ts
```diff
@@ -9,6 +9,7 @@ interface TrackInfo {
 }
 
 interface ThermalRecordingInfo {
+  processingState: string;
```
hardiesoft, line 12: this should be `string | null` right?

test-cypress/cypress/integration/alerts/animal_alerts.ts
```diff
@@ -0,0 +1,78 @@
...
+    cy.apiCreateCamera(cameraTwo, AlertGroup);
+
+    cy.uploadRecording(cameraTwo, {
+      processingState: "FINISHED",
+      tags: ["possum"]
+    });
+    cy.checkAlerts(Tiny, camera, []);
+  });
+
+  it("Alert is based of visit tag, a recording with 2 possum tags and 1 rat is a possum", () => {
```
hardiesoft, line 66: of -> off

hardiesoft (changes requested):

test-cypress/cypress/commands/api/recording.d.ts
gferraro: yep

hardiesoft (approved):

PR #396, opened 2021-07-20
Fix stopped report groupnames

hardiesoft:
This fixes the issue Shaun reported about some groupnames being undefined when there is more than one alert for a group.


hardiesoft:
Actually, never mind, that's a bad PR, device doesn't include groupName, so you'd need another include to pull that in.

PR #398, opened 2021-07-21
Thumbnail

gferraro:
- Read thumnail region from classifier metadata and save in s3 with key  rec.rawFileKey-thumb
- New unauthorized endpoint /api/v1/recordings/:id/thumbnail, returns a png thumbnail if it exists for a recording
- Dont null processingStartTime when it is finished processing.
Added field to recording to indicate if it is in the middle of processing, so we dont pick same recording to process multiple times before it's done once (Instead of checking for a null processingStartTime). Specifically for audio files which go through multiple states before being finished


api/V1/Recording.ts
```diff
@@ -471,6 +475,85 @@ export default (app: Application, baseUrl: string) => {
...
+            "Content-disposition",
+            "attachment; filename=" + filename
+          );
+          response.setHeader("Content-type", mimeType);
+          response.setHeader("Content-Length", data.ContentLength);
+          response.write(data.Body, "binary");
+          return response.end(null, "binary");
+        }
+
+        const range = request.headers.range;
```
hardiesoft: Why would we ever have a range request for a thumbnail image?  We're always going to just server the whole thing, so should never get to here

api/V1/recordingUtil.ts
```diff
@@ -143,6 +145,132 @@ export async function tryToMatchRecordingToStation(
...
+    .promise()
+    .catch((err) => {
+      return err;
+    });
+  //work around for error in cptv-decoder
+  //best to use createReadStream() from s3 when cptv-decoder has support
+  let data = new Uint8Array(fileData.Body);
+  const meta = await decoder.getBytesMetadata(data);
+  let result = await decoder.initWithLocalCptvFile(data);
+  if (!result) {
```
hardiesoft, line 178: You will leak memory here if you exit without doing decoder.close()

api/V1/recordingUtil.ts
```diff
@@ -143,6 +145,132 @@ export async function tryToMatchRecordingToStation(
...
+    "cptv-decoder"
+  );
+  let palette = ColourMaps[0];
+  for (const colourMap of ColourMaps) {
+    if (colourMap[0] == THUMBNAIL_PALETTE) {
+      palette = colourMap;
+    }
+  }
+  renderFrameIntoFrameBuffer(frameBuffer, thumbnail_data, palette[1], 0, 255);
+  const img = await sharp(frameBuffer, {
```
hardiesoft: We only resize if the thumbnail is smaller than the minimum size, but not if it is larger, right?

models/Recording.ts
```diff
@@ -412,7 +412,7 @@ export default function (
           where: {
             type: type,
             processingState: state,
-            processingStartTime: null
+            processing: { [Op.or]: [null, false] }
```
hardiesoft, line 415: Can we make sure that processing is one of the fields returned with a recording by the API?  This will be useful in the UI to show items that are currently processing - at the moment I have an indicator showing that they're queued for processing, but no way of knowing if processing has started yet.

api/V1/recordingUtil.ts
```diff
@@ -143,6 +145,132 @@ export async function tryToMatchRecordingToStation(
...
+  thumbnail.y -= Math.ceil(extra_height);
+  thumbnail.y = Math.max(0, thumbnail.y);
+  thumbnail.height = size;
+  if (thumbnail.y + thumbnail.height > res_y) {
+    thumbnail.y = res_y - thumbnail.height;
+  }
+
+  let frame_start;
+  let thumb_index = 0;
+  for (let i = 0; i < size; i++) {
```
hardiesoft, line 273: I think for a cropped region, you're going to want to get the min/max of the normalisation yourself, since we'd rather normalise for the cropped region than the full frame dynamic range.

api/V1/recordingUtil.ts
```diff
@@ -143,6 +145,132 @@ export async function tryToMatchRecordingToStation(
...
+  let palette = ColourMaps[0];
+  for (const colourMap of ColourMaps) {
+    if (colourMap[0] == THUMBNAIL_PALETTE) {
+      palette = colourMap;
+    }
+  }
+  renderFrameIntoFrameBuffer(frameBuffer, thumbnail_data, palette[1], 0, 255);
+  const img = await sharp(frameBuffer, {
+    raw: { width: thumbnail.width, height: thumbnail.height, channels: 4 }
+  }).resize(size, size, { fit: "contain" });
```
hardiesoft: Maybe it doesn't matter for this use-case, but bear in mind that resizing after doing the palettisation is going to result in colours outside of the palette being created.  So there would be no way of directly mapping the palette back to a greyscale image and remapping to another palette, if we wanted to do that.  Not saying we do, just that it was something we discussed as an idea.

api/V1/recordingUtil.ts
```diff
@@ -143,6 +145,132 @@ export async function tryToMatchRecordingToStation(
...
+  const meta = await img.metadata();
+  const thumbMeta = {
+    width: meta.width.toString(),
+    height: meta.height.toString(),
+    region: JSON.stringify(thumbnail),
+    palette: palette[0]
+  };
+  const data = await img.png().toBuffer();
```
hardiesoft: I think you probably want to use `{ palette: true }` as additional options when creating the png, though it's possible that in some circumstances the rgba default output may be smaller - worth testing.  I also see no reason not to crank the zlib compressionLevel to 9 since it won't be much of a speed hit for such small images: https://sharp.pixelplumbing.com/api-output#png

api/V1/Recording.ts
gferraro: true, i thought it was at the browsers disposition to use range or not. But by not setting accept-ranges we are saying we dont support it

hardiesoft: Yeah, we shouldn't support it for that endpoint, it only makes sense in terms of streaming media formats, of which png is not one.

api/V1/recordingUtil.ts
gferraro: yea i believe thats what "contain" fit does

api/V1/recordingUtil.ts
gferraro: good point mayaswell change it to do palette after resize

api/V1/recordingUtil.ts
gferraro: yep nice catch

models/Recording.ts
gferraro: yep done

hardiesoft:
Moved to cacophony-web

